PR #7, opened 2020-04-17
Added comment in example

kwek20:


kwek20:
Oh wait, i bet its correct when you clone and run the repo, no? :P

Citrullin:
@kwek20 Exactly. With an external import, your imports are correct.

kwek20:
For people new to rust, this message is useful :D

Citrullin:
@kwek20 Probably better to have a full example. Like this one I created. https://github.com/Citrullin/iota-streams-workshop (Not completely working though)
But Jake is working on it. So, you should talk to him.

semenov-vladyslav:
The message is not clear: what exactly is supposed to be replaced with `::`?

The issue here is with crates naming, ie. `iota_streams` crate just reexports all other subcrates with shorter names. You can either use `iota_streams` or use specific `iota_streams_xxx` crates. A possible solution would be to move `basic_scenario.rs` into `iota_streams` examples, but it's just a literally basic scenario of using `iota_streams_app_channels` and it wouldn't make sense to move it anywhere.

Please, refer to [rust book](https://doc.rust-lang.org/book/ch07-00-managing-growing-projects-with-packages-crates-and-modules.html) for explanation.

PR #36, opened 2020-09-26
Merge user and transport instances

DyrellC:
This PR is to introduce a merging of transport and user api's to reduce the number of steps needed by the users to utilise Streams functionality. It takes the transport layer and pulls it into the User object so that users only need to use one set of API tools rather than individual Transport and User instances:

- All send and receive functions are enacted through the wrapping/unwrapping api
- Links and sequence states will only be stored if a message is properly sent through the transport layer
- Added Rc to core::prelude 
- Author and Subscriber API's updated to use links instead of messages
- Moved fetching logic from examples utility to tangle API instances
- Examples updated with new api arguments


iota-streams-app-channels/src/api/mod.rs
```diff
@@ -145,6 +149,34 @@ pub mod user;
 #[cfg(all(feature = "tangle"))]
 impl<F> ChannelLinkGenerator<TangleAddress> for DefaultTangleLinkGenerator<F> where F: PRP {}
 
+/// Message associated info, just message type indicator.
+#[derive(Copy, Clone)]
+pub enum MsgInfo {
```
semenov-vladyslav, line 154: This stuff is really tangle specific. Message info is some auxiliary data associated to a link stored in `link_store`. It allows to keep a list of recipients for Keyload message, signer public key for SignedPacket or app-specific data. `MsgInfo` is just an enum because it's difficult to think of a good enough generic type for channels app.

iota-streams-app-channels/src/api/tangle/author.rs
```diff
@@ -4,38 +4,59 @@ use anyhow::{
...
 
 use iota_streams_core::{
-    prelude::Vec,
+    prelude::{Vec, Rc},
     prng,
 };
 use iota_streams_core_edsig::signature::ed25519;
 
-type AuthorImp = User<DefaultF, Address, LinkGen, LinkStore, PkStore, PskStore>;
+type AuthorImp<T> = User<DefaultF, Address, LinkGen, LinkStore, PkStore, PskStore, T>;
```
semenov-vladyslav, line 32: Shall we introduce some default transport? How otherwise do we implement c bindings?

iota-streams-app-channels/src/api/tangle/author.rs
```diff
@@ -164,41 +169,56 @@ impl Author {
...
-    // ids.push((pk.0, msgid1, seq_num1));
-    // }
-    // }
-    // ids
-    // }
     pub fn gen_next_msg_ids(&mut self, branching: bool) -> Vec<(ed25519::PublicKey, SequencingState<Address>)> {
         self.imp.gen_next_msg_ids(branching)
     }
+
+    pub fn fetch_next_msgs(&mut self) -> Result<Vec<(ed25519::PublicKey, Address, Bytes, Bytes)>> {
```
semenov-vladyslav, line 176: What is the return type?

iota-streams-app-channels/src/api/tangle/author.rs
```diff
@@ -164,41 +169,56 @@ impl Author {
...
+        let ids = self.gen_next_msg_ids(self.is_multi_branching());
+        let mut msgs = Vec::new();
+        let mut exists = false;
+
+        for (pk, SequencingState(link, seq)) in ids {
+            let msg = (&*self.imp.transport).borrow_mut().recv_message_with_options(&link, &self.imp.recv_opt);
```
semenov-vladyslav, line 182: `msg` is of `Result` type, maybe there's a better name?

iota-streams-app-channels/src/api/tangle/author.rs
```diff
@@ -164,41 +169,56 @@ impl Author {
...
+            if msg.is_ok() {
```
semenov-vladyslav, line 183: Do we just skip the error as there's no else branch? Why?

iota-streams-app-channels/src/api/tangle/author.rs
```diff
@@ -164,41 +169,56 @@ impl Author {
...
+                            msgs.push((pk, next_link.clone(), content.0, content.1));
+                            break;
+                        }
+                        message::SEQUENCE => {
+                            let msg_link = self.unwrap_sequence(next_link.clone())?;
+                            msg = (&*self.imp.transport).borrow_mut().recv_message_with_options(&msg_link, &self.imp.recv_opt)?;
+                            self.store_state(pk.clone(), next_link.clone());
+                            next_link = msg_link;
+                        }
+                        _ => {
```
semenov-vladyslav, line 208: We don't handle other types of messages? Keyload for example?

iota-streams-app-channels/src/api/tangle/author.rs
```diff
@@ -164,41 +169,56 @@ impl Author {
...
+            let msg = (&*self.imp.transport).borrow_mut().recv_message_with_options(&link, &self.imp.recv_opt);
+            if msg.is_ok() {
+                let mut msg = msg.unwrap();
+                let mut next_link = link.clone();
+
+                loop {
+                    let preparsed = msg.parse_header()?;
+                    match preparsed.header.content_type {
+                        message::SIGNED_PACKET => {
+                            let content = self.unwrap_signed_packet(next_link.clone())?;
```
semenov-vladyslav, line 191: If unwrap fails here for some reason we are going to lose all `msgs` handled by this moment. Is that ok?

examples/src/branching/utils.rs
```diff
@@ -80,62 +32,24 @@ pub fn s_fetch_next_messages<T: Transport>(
...
-) where
+pub fn a_fetch_next_messages<T: Transport>(author: &mut Author<T>)
+where
     T::RecvOptions: Copy,
 {
     let mut exists = true;
 
     while exists {
-        let ids = author.gen_next_msg_ids(multi_branching);
+        let msgs = author.fetch_next_msgs().unwrap();
```
semenov-vladyslav, line 42: `unwrap` will `panic` if there's an error.

semenov-vladyslav (changes requested):
There's a few issues that need to be addressed:
- Transport: Rc/RefCell, async, Options, impl for c bindings
- logic behind fetch_next_msgs and MsgInfo type

iota-streams-app-channels/src/api/user.rs
```diff
@@ -76,9 +85,15 @@ pub struct User<F, Link, LG, LS, PKS, PSKS> {
     pub message_encoding: Vec<u8>,
 
     pub uniform_payload_length: usize,
+
+    pub(crate) transport: Rc<RefCell<T>>,
```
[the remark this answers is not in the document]
semenov-vladyslav: I actually think that it's best to split `User` into tangle `User` part that will contain api `User` and transport and will handle sending/recving messages. And we can maybe get rid off of `handle_xxx` methods from api `User`.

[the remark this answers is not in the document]
semenov-vladyslav: Also, should we implement transport as `Box<dyn Transport>` to hide transport type from type signature and allow transport polymorphism?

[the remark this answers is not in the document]
DyrellC: To address the top part we'll be doing the Default Transport implementation, so we should be able to avoid the need to actually use Rc and RefCell altogether. Those are only necessary fro the example purposes and don't really make sense in a production environment. Brord should be tackling that right now. 

As for the second point, we'll be moving the Transport up into the Tangle layer with that new User implementation wrapping the Auth/Sub instance with the transport and a simplified/symmetric api (Hurray).

iota-streams-app-channels/src/api/user.rs
```diff
@@ -76,9 +85,15 @@ pub struct User<F, Link, LG, LS, PKS, PSKS> {
...
+
+    pub(crate) recv_opt: <T as transport::Transport<F, Link>>::RecvOptions,
```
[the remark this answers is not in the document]
DyrellC: Brord should be on top of this one, but we'll be using this approach to set and get the recv and send options. They should (in most implementations) just be the defaults, but for those using private tangles with a reduced mwm for example will need the ability to set what those "defaults" would actually be.

iota-streams-app-channels/src/api/user.rs
```diff
@@ -165,6 +187,7 @@ where
         info: <LS as LinkStore<F, <Link as HasLink>::Rel>>::Info,
     ) -> Result<BinaryMessage<F, Link>> {
```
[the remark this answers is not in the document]
DyrellC: This, and others have been set to return links now.

iota-streams-app-channels/src/api/user.rs
```diff
@@ -165,6 +187,7 @@ where
...
         let wrapped = self.prepare_announcement()?.wrap()?;
+        (&*self.transport).borrow_mut().send_message_with_options(&wrapped.message, &self.send_opt)?;
```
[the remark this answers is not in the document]
DyrellC: We'll have to address async post release but it is indeed a thing we need to address in the future. For now it should be noted that Users are expected to track their message sent status themselves for async. Handling sync states may need to be made into a callback to be handled on a future completion, but we'll address that later.

iota-streams-app-channels/src/api/user.rs
```diff
@@ -191,9 +214,12 @@ where
     /// in the message.
     pub fn handle_announcement<'a>(
         &mut self,
-        preparsed: PreparsedMessage<'a, F, Link>,
+        link: Link,
```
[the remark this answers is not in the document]
DyrellC: This problem is essentially present everywhere not just with announce. We now will have a handle message function that will automate this process though.

iota-streams-app/src/transport/mod.rs
```diff
@@ -55,18 +55,19 @@ pub trait Transport<F, Link> // where Link: HasLink
...
     fn recv_message(&mut self, link: &Link) -> Result<BinaryMessage<F, Link>>
     where
         Self::RecvOptions: Default,
     {
-        self.recv_message_with_options(link, Self::RecvOptions::default())
+        self.recv_message_with_options(link, &Self::RecvOptions::default())
     }
 }
 
+#[derive(Clone)]
```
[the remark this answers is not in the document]
DyrellC: Removed

iota-streams-app-channels/src/api/mod.rs
DyrellC: This is moved up into the tangle layer again

iota-streams-app-channels/src/api/tangle/user.rs
```diff
@@ -0,0 +1,302 @@
...
+
+pub trait SubUser: UserImp {
+    fn is_registered(&self) -> bool;
+    fn unregister(&mut self);
+    fn subscribe(&mut self, link_to: &Address) -> Result<WrappedMessage>;
+    fn unwrap_announcement<'a>(&mut self, msg: Message) -> Result<()>;
+}
+
+
+pub struct User<Trans, U>
```
semenov-vladyslav, line 57: Where is this `User` going to be used? Does `U` stand for `Author` or `Subscriber` from `tangle` module? Is it going to be used as a top entity and `Author` and `Subscriber` become implementation details?

semenov-vladyslav: I had a slightly different idea instead:
```
mod api { struct User }
mod tandle {
pub struct User<Tsp> {
    tsp: Tsp,
    user: api::User,
}
pub struct Author<Tsp> {
    user: User<Tsp>, // NOTE: User from tangle mod, not api
}
pub struct Subscriber<Tsp> {
    user: User<Tsp>, // NOTE: User from tangle mod, not api
}
}
```
This way `User`, `Author` and `Subscriber` stay the top-level entities, and we can nicely implement `IsUser`, `IsAuthor` and `IsSubscriber` traits.

semenov-vladyslav (changes requested):
Most importantly deal with `User` struct, the current design is dubious. Also need to deal with `Result` and not just skip errors.

iota-streams-app-channels/src/api/tangle/user.rs
DyrellC: So the main reason why there were the traits was to provide functionality that works between user types so we don't have to repeat everything. When we remove the traits and do struct implementations like this we get locked up into having to implement everything individually on each. Making the api a little more verbose than necessary. There is no api::User to pull in for example, we can only pull in api::user::User, which is the api implementation, not a superclass of Author and Sub, so because of that we would be directly injecting the Author/Sub implementation into the User rather than using the Tangle api layer at all. If user::User was a trait being applied then it would maybe make sense. The idea of the implementation I wrote up here was to act as a Top entity that takes in either an Author or Subscriber, and the Auth and Subs use the traits to use the base layer functionality present in the tangle::User class.

User gets used as the baseline implementation when setting up your instance, and depending on whether you load in an Author or Subscriber implementation you will be restricted to using the functions present in the traits for those specifically. So it also provided the separation of powers for functionality. A User with an Author injected can use the functions present in the UserImp trait, but also the AuthUser trait, but not the SubUser trait. 

As is I don't really see how to apply it the way you want here because there is no "inheritable" trait to apply these structs to that wouldn't require us to rewrite how the Author and Sub are used right now.

iota-streams-app-channels/src/api/tangle/user.rs
```diff
@@ -0,0 +1,302 @@
...
+                message::TAGGED_PACKET => {
+                    let content = self.user.unwrap_tagged_packet(msg);
+                    if content.is_ok() {
+                        let content = content.unwrap();
+                        unwrapped_content = Some((pk, next_link.clone(), content.0, content.1));
+                    }
+                    break;
+                }
+                message::KEYLOAD => {
+                    let _unwrapped = self.user.unwrap_keyload(msg);
```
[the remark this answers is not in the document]
DyrellC: This is checked inside the api user now, it'll return a bool if you are part of the keyload or not. If a different failure occurs then it will fail out (with the change from Option to Result for the return).

semenov-vladyslav (approved):

PR #37, opened 2020-10-02
Made example load optionally from env

kwek20:
Env loading does have a STD requirement, do we want this? I think it helps :P 
I keep forgetting to switch back to localhost from my node (which isnt local)

Todo: 
- [x] Add env to bindings

### DIFF
--- a/bindings/c/main.c
+++ b/bindings/c/main.c
@@ -28,12 +28,18 @@ int main()
   rand_seed(seed, sizeof(seed));
 
 #ifdef IOTA_STREAMS_CHANNELS_CLIENT
-  char url[] = "http://localhost:14265";
+  char *env_url = getenv("URL");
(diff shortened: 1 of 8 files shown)


DyrellC (approved):
I like it.

bindings/c/main.c
```diff
@@ -28,12 +28,18 @@ int main()
   rand_seed(seed, sizeof(seed));
 
 #ifdef IOTA_STREAMS_CHANNELS_CLIENT
-  char url[] = "http://localhost:14265";
+  char *env_url = getenv("URL");
```
semenov-vladyslav, line 31: `char const *` - should not be able to modify that string.

semenov-vladyslav (approved):
Looks good but we can change a few `match`es with `map`s.

PR #47, opened 2020-10-19
Removed conversion dep

kwek20:
We still had this old guy in there, no more!


DyrellC (approved):
Lovely, one less dependency to build.

PR #49, opened 2020-10-19
Updated docs to 1.3

Tsangares:
Small documentation change. The documentation was linking to 1.2 instead of 1.3.

A permanent fix would be if the url `https://docs.iota.org/docs/channels/` would link to the latest documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ fn main() {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


kwek20:
Great idea; i've opened an issue! https://github.com/iotaledger/documentation/issues/822

PR #50, opened 2020-11-09
Update c bindings, add new functions

DyrellC:
# Description of change
Updating c bindings to include tagged and signed packets on the subscriber end, as well as a couple utility functions for usability (like converting string link representation to proper streams Address object through bindings).

## Type of change

I have run each function through modifications of the main.c file, and confirmed they work when pulled into golang

## Change checklist


bindings/c/src/api/mod.rs
```diff
@@ -23,6 +23,24 @@ use cty::{
 
 use iota::client::bytes_to_trytes;
 
+#[no_mangle]
+pub extern "C" fn address_from_string(c_addr: *const c_char) -> *const Address {
+    unsafe {
+        let addr_vec: Vec<&str> = CStr::from_ptr(c_addr).to_str().unwrap().split(":").collect();
+        let addr = Address::from_str(addr_vec[0], addr_vec[1]).unwrap();
```
semenov-vladyslav, line 30: `addr_vec[1]` will panic if address doesn't contain a colon, need to add a guard. 

`unwrap()` in general can panic, need better handling of unwraps throughout all streams repo, not only bindings. In c bindings we can return nulls.

bindings/c/src/api/mod.rs
```diff
@@ -136,6 +154,11 @@ impl Default for MessageLinks {
     }
 }
 
+#[no_mangle]
+pub extern "C" fn new_message_links(msg_link: *const Address, seq_link: *const Address) -> *mut MessageLinks {
```
semenov-vladyslav, line 158: No need to make an explicit function, `message_links_t` struct is defined in `channels.h` and can be constructed directly in c.

bindings/c/src/api/mod.rs
```diff
@@ -269,19 +292,34 @@ pub extern "C" fn get_msgid_str(msgid: *mut MsgId) -> *const c_char {
...
     unsafe {
         CString::new(bytes_to_trytes((*address).msgid.as_ref()).to_string()).unwrap().into_raw()
     }
 }
 
+
+#[no_mangle]
+pub extern "C" fn get_address_id_str(address: *mut Address) -> *mut c_char {
+    unsafe {
+        CString::new(hex::encode((*address).msgid.as_ref())).unwrap().into_raw()
```
semenov-vladyslav, line 319: Shall we check pointers for `null`, or hope/expect caller code doesn't violate convention?

semenov-vladyslav (changes requested):
Update `main.c` to cover the new functions.

bindings/c/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ include(FetchContent)
 
 option(NO_STD "Enable no_std build, without iota_client" OFF)
 option(SYNC_CLIENT "Enable sync transport via iota_client" ON)
-option(STATIC "Build shared library" ON)
+option(STATIC "Build shared library" OFF)
```
kwek20, line 8: ahyes we should rename this and maybe decide on what we make as default, we keep swapping it aboot

bindings/c/src/api/mod.rs
DyrellC: Hmmm a good question really. I think it's fair to expect that they'll not violate the conventions, but it probably doesn't hurt to catch the strays and save them the seg fault

bindings/c/src/api/mod.rs
DyrellC: Will map_or them to ensure null returns wherever a panic could occur

bindings/c/CMakeLists.txt
DyrellC: We should default to Shared since static is still giving us grief

kwek20 (approved):
LGTM!

bindings/c/src/api/mod.rs
```diff
@@ -21,7 +21,29 @@ use cty::{
     uint8_t,
 };
 
-use iota::client::bytes_to_trytes;
+
+#[no_mangle]
+pub extern "C" fn address_from_string(c_addr: *const c_char) -> *const Address {
```
semenov-vladyslav, line 26: This is a nice function in general, it's going to be used in other bindings. Maybe move the impl to the main crate?

semenov-vladyslav (changes requested):
Looks good, but move impl of `address_from_string` into main crate?

bindings/c/src/api/mod.rs
DyrellC: Yeah I think that's a good idea, i'll do that before push

PR #95, opened 2021-05-31
Update develop to c2 

DyrellC:
# Description of change
Bring Develop up to speed with chrysalis branch. We'll direct changes to develop from now on to avoid breaking any implementations built on c2 before giving out warnings for them.


kwek20 (approved):
Totally mint

PR #97, opened 2021-06-09
Update Documentation

DyrellC:
# Description of change
Updating the rust docs for Chrysalis 2. This will include:
- [x] Updates to the rust books README's 
- [x] Updated API reference
- [x] Updated c Binding API reference
- [x] Updated WASM Binding API reference 

Additional examples: https://github.com/iotaledger/streams-examples

## Type of change
- Documentation Fix

N/A

## Change checklist


dominic22:
I had a look at the examples and would have expected more detailed ones. Here are examples we've discussed during the last weeks:
- multi branching in detail (two subscribers, one author, one reader from beginning, one reader in the middle of channel, one auditor added when all data on channel is written and can read using a preshared key)
- public channel in single branch
- public channel in multi branch
- shared key for single branch
- shared key for multi branch (might be obsolete with multi branch example described above)
- using permanode for readers of a channel

Something not discussed but would be from interest:
- analyze message types: for instance analyze as author if new subscription to channel is available and trigger a basic console.log if one is available

docs/libraries/wasm/examples.md
```diff
@@ -1 +1,123 @@
...
+```
+
+### Message Fetching 
+#### Forward
+When new messages are available to retrieve from the channel, you can fetch the next 
+message sent by each publisher like so:
+```
+let next_msgs = sub.clone().fetch_next_msgs();
+
+for (var i = 0; i < prev_msgs.length; i++) {
```
sobolev-alexey, line 94: I'd assume that you want to iterate over `next_msgs`, not prev_msgs

docs/libraries/wasm/examples.md
```diff
@@ -1 +1,123 @@
...
+      from_bytes(prev_msgs[i].get_message().get_public_payload()),
+      "\tMasked: ",
+      from_bytes(prev_msgs[i].get_message().get_masked_payload())
+    );
+}
+```
+
+If no new messages are present, the returned array will be empty.
+
+You can also fetch all possible next messages
```
sobolev-alexey, line 107: You can also fetch ~~all possible next messages~~ **prev** messages

sobolev-alexey:
I agree with @dominic22 
we see a good progress here, however some of the important parts are still missing. 

In addition to the list Dominic posted above I would also add an option to read the Stream (forwards/backwards) from any defined position within the stream. 
This can be done by e.g. providing an index as parameter to the `fetch_next_msgs()` and `fetch_prev_msgs()` functions

sobolev-alexey:
Some additional examples can be found in [this repository](https://github.com/iotaledger/streams-examples)

PR #98, opened 2021-06-09
WASM: Add Stateless Recovery

DyrellC:
# Description of change
Add recover functions to Wasm Bindings. Required for stateless recovery of a user instance from the WASM end. It emulates MAM like behaviour

## Type of change

Examples have been updated to include functionality for testing

## Change checklist


bindings/wasm/src/lib.rs
```diff
@@ -22,6 +22,17 @@ extern "C" {
     pub fn error(s: &str);
 }
 
+pub async fn wait(ms: u32) {
```
kwek20, line 25: Replace this with `await new Promise(r => setTimeout(r, ms));`

kwek20 (changes requested):
We should really change it so that the user doesnt need to call .clone() on auth and sub. We can do this ourselves inside the methods right? Or am i missing something

bindings/wasm/src/lib.rs
DyrellC: I tried to do that first and ran into a whole host of issues trying to import the function. In order to use `setTimeout()` we need it to build the js function in commonjs not ES6, which wasm-bindgen doesn't support (see https://rustwasm.github.io/wasm-bindgen/reference/js-snippets.html#caveats). So building this for Web target is fine, but not for nodejs. We could look at using the Closures instead maybe? https://rustwasm.github.io/wasm-bindgen/reference/passing-rust-closures-to-js.html

DyrellC: I'm not sure we can avoid the `clone()` either, since the js functions consume the author or subscriber instance given to them (or so is my understanding) so the `clone()` is done to pass a new instance of Author/Subscriber with a copy of the `Rc<RefCell<ApiAuthor/ApiSubscriber>>`. If you try to pass by reference instead of by static object the compiler complains that `me` doesn't live long enough.

kwek20: @DyrellC I figured out a way, can you confirm they work for you as well?
I can both `build:nodejs` and `build:web`

DyrellC:
@kwek20  The `.clone()` removal is in https://github.com/iotaledger/streams/pull/140, and i've removed the listen function, now it's just stateless recovery, and it looks like it's good to go as soon as it's approved here.

bindings/wasm/src/author/authorw.rs
```diff
@@ -79,6 +79,25 @@ impl Author {
         block_on(self.author.borrow_mut().export(password)).into_js_result()
     }
 
+    #[wasm_bindgen(catch)]
```
arnauorriols, line 82: unnecessary

bindings/wasm/src/subscriber/subscriberw.rs
```diff
@@ -64,6 +64,20 @@ impl Subscriber {
             .into_js_result()
     }
 
+    #[wasm_bindgen(catch)]
```
arnauorriols, line 67: idem

bindings/wasm/examples/node.ts
```diff
@@ -147,6 +147,18 @@ async function main() {
         console.log("import succesfull")
     }
 
+<<<<<<< HEAD
```
arnauorriols, line 150: Fix conflict

DyrellC: Doing that now actually, caught it right after i pushed :sweat_smile:

bindings/wasm/src/author/authorw.rs
DyrellC: accurate

arnauorriols (commented):
Good to go once you fix the git conflict!

arnauorriols (approved):

DyrellC:
Alright conflict fixed up and attributes removed @arnauorriols

PR #101, opened 2021-06-28
TransportDetails using trait

kwek20:
# Description of change

Adds getting link details from the specific transport type

Example:
```rust
    let node_url = env::var("URL").unwrap_or_else(|_| "https://chrysalis-nodes.iota.org".to_string());

    let mut client = Client::new_from_url(&node_url);
    let mut author = Author::new(....);

    println!("\nAnnounce Channel");
    let announcement_link = author.send_announce()?;

    let details = client.get_link_details(&announcement_link).unwrap();
    dbg!(details.milestone);
```

## Type of change

Choose a type of change, and delete any options that are not relevant.

Ran it... it shows details :D

## Change checklist

- [X] I have followed the contribution guidelines for this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes


semenov-vladyslav (approved):
Looks good.

C bindings expose structs for Client transport details. The struct layout is a subject for changes. One limitation is due to a fixed number of parent_message_ids (which is fixed to be 2 and can be bigger in parctice), this issue will be addressed later. Such approach is simpler as it gives access to a binary struct with one function. Another approach would be to make Client transport details struct abstract and only accessible via getters/setters. For the moment the latter approach seems unnecessary complicated as it bloats bindings boilerplate code.

PR #113, opened 2021-07-14
Feature reset state

DyrellC:
# Description of change
Add `reset_state` function to subscribers to allow subscribers to fetch messages from scratch without restarting their instance. 

Example: 
```
subscriber.reset_state()?;
let msgs = subscriber.fetch_all_next_msgs();
```

Bindings and  docs have been updated as well. 

## Type of change

New section in examples demonstrating usage in rust, c and wasm.

## Change checklist

### DIFF
--- a/docs/libraries/wasm/api_reference.md
+++ b/docs/libraries/wasm/api_reference.md
@@ -76,6 +76,12 @@ Retrieve the Author public key.
 | --------------- | ------------------- | ------------------------- |
 **Returns:** The Author public key in hex representation.
 
+#### fetch_state(): Array<[UserState](#UserState)>
+Retrieve the currently known publisher states for the channel.
+
+| Param           | Type                | Description               |
+| --------------- | ------------------- | ------------------------- |
+**Returns:** An array of user state objects representing the currently known state of all publishers in the channel.
 
 
 #### The following functions require author.clone() to use, as they consume the instance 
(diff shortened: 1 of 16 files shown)


iota-streams-app-channels/src/api/user.rs
```diff
@@ -253,6 +253,23 @@ where
...
 
+    /// Reset link store and key store to original state
+    pub fn reset_state(&mut self) -> Result<()> {
+        match &self.appinst {
+            Some(appinst) => {
+                let mut pk_store = PKS::default();
+                for (pk, _cursor) in self.pk_store.iter() {
+                    pk_store.insert(*pk, Cursor::new_at(appinst.rel().clone(), 0, 2_u32))?;
+                }
+                self.pk_store = pk_store;
```
semenov-vladyslav, line 264: Looks like a dirty hack. Currently, User state contains two kinds of objects: permanent - keys, settings, etc., and ephemeral - current index within channel (ie. cursor for all known users). This index is stored in pk_store that serves these two purposes: it contains public keys and associated cursor. It begs to be split in two: just public keys and such index. This index then can be made an external object that can be manipulated more easily.

DyrellC: What would be your proposition for resetting state then if not to literally reset it in this fashion? The pk_store needs to be put back to the starting state in order to facilitate syncing from scratch, which means the cursors need to be reset to the starting point for all publishers. Technically it could be done by only including the author pk and setting it the same way receiving an announcement would, and once the subscriber hits the keyloads they'll re-enter any other publishers in the standard way, but that just postpones the placement of those pk/cursor mappings until then, it doesn't actually change the behaviour at all?

docs/libraries/wasm/api_reference.md
```diff
@@ -655,3 +676,40 @@ Fetch Public Keys in string formatting
...
+**Returns:** The latest published message link 
+
+#### get_seq_no()
+Get the sequence state number from the internal cursor object 
+
+| Param     | Type                  | Description                                  |
+| --------- | --------------------- | -------------------------------------------- |
+**Returns:** The latest sequence state number of the publisher 
+
+#### get_seq_no()
```
kwek20, line 710: typo here, still says seq

kwek20 (changes requested):

docs/libraries/wasm/api_reference.md
DyrellC: Tis done

PR #116, opened 2021-07-15
Added Clone trait to transport

kwek20:
Added Clone so we can use Transport outside of Streams repo without specifying Clone all over the place!


DyrellC (approved):
Definitely a welcome change imo. Lgtm

PR #132, opened 2021-08-02
Re-enable some lints from app and app-channels

arnauorriols:
# Description of change

Some Clippy lints were disabled temporarily to favor speed of delivery of needed features. This PR refactors the several parts of the code to re-enable some of those lints.   

## Type of change

Specially relevant breaking changes:
- Messages are now indexed using the binary digest of the `Address` blake2b256 hash. Messages were previously hashed using the hexadecimal encoding of said hash. Warning: old messages will **not** be compatible with this new version.
- `Address` `Display` format now outputs the same as `Address::to_string()`. To get the hash of the address one must use `Address::to_blake2b()` or `Address::to_msg_index()`

Run the following examples:

- streams/examples
- bindings/wasm examples
- bindings/c examples
- iota-streams-app-channels-example

## Change checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,7 @@
 [package]
 name = "iota-streams"
 version = "1.0.0"
-authors = ["Vlad Semenov <vlad.semenov@iota.org>", "Dyrell Chapman <dyrell.chapman@iota.org>", "Brord van Wierst <brord@iota.org>"]
+authors = ["Vlad Semenov <vlad.semenov@iota.org>", "Dyrell Chapman <dyrell.chapman@iota.org>", "Brord van Wierst <brord@iota.org>", "Arnau Orriols <arnau.orriols@iota.org"]
 edition = "2018"
 license = "Apache-2.0/MIT"
 readme = "README.md"
--- a/bindings/c/CMakeLists.txt
+++ b/bindings/c/CMakeLists.txt
@@ -19,7 +19,7 @@ if(${SYNC_CLIENT})
   set(cargo_features "${cargo_features}sync-client")
 endif(${SYNC_CLIENT})
 
-message("NO_STD=${NO_STD} SYNC_CLIENT=${SYNC_CLIENT} STATIC=${STATIC}")
+message("NO_STD=${NO_STD} SYNC_CLIENT=${SYNC_CLIENT} STATIC=${STATIC} RELEASE=${RELEASE}")
 
 include_directories(include/)
 
--- a/bindings/c/CMakeLists.txt
+++ b/bindings/c/CMakeLists.txt
@@ -4,8 +4,8 @@ project(iota_streams_c LANGUAGES C)
 include(FetchContent)
 
 option(NO_STD "Enable no_std build, without iota_client" OFF)
-option(SYNC_CLIENT "Enable sync transport via iota_client" OFF)
-option(STATIC "Build static library" ON)
+option(SYNC_CLIENT "Enable sync transport via iota_client" ON)
+option(STATIC "Build static library" OFF)
 option(RELEASE "Build release library (defaults to release)" ON)
 
 set(cargo_features "")
--- a/bindings/c/src/api/auth.rs
+++ b/bindings/c/src/api/auth.rs
@@ -171,8 +171,10 @@ pub unsafe extern "C" fn auth_send_keyload(
             link_to.as_ref().map_or(Err::NullArgument, |link_to| {
                 psk_ids.as_ref().map_or(Err::NullArgument, |psk_ids| {
                     ke_pks.as_ref().map_or(Err::NullArgument, |ke_pks| {
-                        let identifiers: Vec<Identifier> = ke_pks.into_iter().map(|pk| (*pk).into()).collect();
-                        user.send_keyload(link_to, psk_ids, &identifiers.iter().collect())
+                        let pks = ke_pks.into_iter().copied().map(Into::<Identifier>::into);
+                        let psks = psk_ids.into_iter().copied().map(Into::<Identifier>::into);
+                        let identifiers: Vec<Identifier> = pks.chain(psks).collect();
+                        user.send_keyload(link_to, &identifiers)
                             .map_or(Err::OperationFailed, |response| {
                                 *r = response.into();
                                 Err::Ok
--- a/bindings/c/src/api/mod.rs
+++ b/bindings/c/src/api/mod.rs
@@ -11,9 +11,7 @@ use iota_streams::{
         },
         identifier::Identifier,
         message::Cursor,
-        transport::tangle::{
-            MsgId,
-        },
+        transport::tangle::MsgId,
     },
     app_channels::api::{
         psk_from_seed,
@@ -28,6 +26,7 @@ use iota_streams::{
 
 use core::ptr::{
     null,
+    null_mut,
 };
 
 pub fn get_channel_type(channel_type: uint8_t) -> ChannelType {
@@ -59,6 +58,24 @@ pub(crate) fn safe_drop_mut_ptr<T>(p: *mut T) {
     }
 }
 
+/// Convert an String-like collection of bytes into a raw pointer to the first byte
+///
+/// The pointer might be [`null`] if the String contains a null byte (which is invalid)
+///
+/// [`null`]: https://doc.rust-lang.org/std/ptr/fn.null.html
+fn _string_into_raw(string: impl Into<Vec<u8>>) -> *const c_char {
+    CString::new(string).map_or_else(|_e| null_mut(), CString::into_raw)
+}
+
+/// Convert an String-like collection of bytes into a raw pointer to the first byte
+///
+/// This function is unsafe because it does not check that the String does not contain a null byte.
+/// Use this function instead of [`string_into_raw`] in those cases where it's certain there won't be
+/// a null byte and don't want to incur the performance penalty of the validation.
+unsafe fn string_into_raw_unchecked(string: impl Into<Vec<u8>>) -> *const c_char {
+    CString::from_vec_unchecked(string.into()).into_raw()
+}
+
 #[repr(C)]
 pub enum Err {
     Ok,
@@ -74,9 +91,9 @@ pub unsafe extern "C" fn address_from_string(c_addr: *const c_char) -> *const Ad
 
 #[no_mangle]
 pub unsafe extern "C" fn public_key_to_string(pubkey: *const PublicKey) -> *const c_char {
-    pubkey.as_ref().map_or(null(), |pk| {
-        CString::new(hex::encode(pk.as_bytes())).map_or(null(), |pk| pk.into_raw())
-    })
+    pubkey
+        .as_ref()
+        .map_or(null(), |pk| string_into_raw_unchecked(hex::encode(pk.as_bytes())))
 }
 
 #[no_mangle]
@@ -89,9 +106,9 @@ pub type KePks = Vec<PublicKey>;
 
 #[no_mangle]
 pub unsafe extern "C" fn pskid_as_str(pskid: *const PskId) -> *const c_char {
-    pskid.as_ref().map_or(null(), |pskid| {
-        CString::new(hex::encode(&pskid)).map_or(null(), |id| id.into_raw())
-    })
+    pskid
+        .as_ref()
+        .map_or(null(), |pskid| string_into_raw_unchecked(hex::encode(&pskid)))
 }
 
 #[no_mangle]
@@ -528,29 +545,29 @@ pub unsafe extern "C" fn drop_str(string: *const c_char) {
 #[no_mangle]
 pub unsafe extern "C" fn get_channel_address_str(appinst: *const ChannelAddress) -> *const c_char {
     appinst.as_ref().map_or(null(), |inst| {
-        CString::new(inst.to_string()).map_or(null(), |inst_str| inst_str.into_raw())
+        // Calling `to_hex_string()` instead of `to_string()` certifies that the String won't contain
+        // a null byte, so that we can call `string_into_raw_unchecked()`
+        string_into_raw_unchecked(inst.to_hex_string())
     })
 }
 
 #[no_mangle]
 pub unsafe extern "C" fn get_msgid_str(msgid: *const MsgId) -> *const c_char {
-    msgid.as_ref().map_or(null(), |id| {
-        CString::new(id.to_string()).map_or(null(), |id_str| id_str.into_raw())
-    })
+    msgid
+        .as_ref()
+        .map_or(null(), |id| string_into_raw_unchecked(id.to_hex_string()))
 }
 
 #[no_mangle]
 pub unsafe extern "C" fn get_address_inst_str(address: *const Address) -> *const c_char {
-    address.as_ref().map_or(null(), |addr| {
-        get_channel_address_str(&addr.appinst)
-    })
+    address
+        .as_ref()
+        .map_or(null(), |addr| get_channel_address_str(&addr.appinst))
 }
 
 #[no_mangle]
 pub unsafe extern "C" fn get_address_id_str(address: *const Address) -> *const c_char {
-    address.as_ref().map_or(null(), |addr| {
-        get_msgid_str(&addr.msgid)
-    })
+    address.as_ref().map_or(null(), |addr| get_msgid_str(&addr.msgid))
 }
 
 #[no_mangle]
@@ -558,7 +575,7 @@ pub unsafe extern "C" fn get_address_index_str(address: *const Address) -> *cons
     address.as_ref().map_or(null(), |addr| {
         let index = addr.to_msg_index();
         let index_hex = format!("{:x}", index);
-        CString::new(index_hex).map_or(null(), |index| index.into_raw())
+        string_into_raw_unchecked(index_hex)
     })
 }
 
--- a/bindings/c/src/api/mod.rs
+++ b/bindings/c/src/api/mod.rs
@@ -368,7 +368,6 @@ impl MessageLinks {
         safe_drop_ptr(self.msg_link);
         safe_drop_ptr(self.seq_link);
     }
-
 }
 
 impl Default for MessageLinks {
@@ -394,8 +393,6 @@ pub unsafe extern "C" fn get_msg_link(msg_links: *const MessageLinks) -> *const 
 pub unsafe extern "C" fn get_seq_link(msg_links: *const MessageLinks) -> *const Address {
     msg_links.as_ref().map_or(null(), |links| links.seq_link)
 }
-
-
 
 #[repr(C)]
 pub struct Buffer {
--- a/bindings/c/src/api/mod.rs
+++ b/bindings/c/src/api/mod.rs
@@ -521,7 +521,7 @@ pub extern "C" fn drop_payloads(payloads: PacketPayloads) {
 }
 
 #[no_mangle]
-pub unsafe extern "C" fn drop_str(string: *mut c_char) {
+pub unsafe extern "C" fn drop_str(string: *const c_char) {
     CString::from_raw(string as *mut c_char);
 }
 
--- a/bindings/c/src/api/mod.rs
+++ b/bindings/c/src/api/mod.rs
@@ -12,7 +12,6 @@ use iota_streams::{
         identifier::Identifier,
         message::Cursor,
         transport::tangle::{
-            get_hash,
             MsgId,
         },
     },
@@ -29,7 +28,6 @@ use iota_streams::{
 
 use core::ptr::{
     null,
-    null_mut,
 };
 
 pub fn get_channel_type(channel_type: uint8_t) -> ChannelType {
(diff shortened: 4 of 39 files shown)


Cargo.toml
```diff
@@ -3,7 +3,7 @@
 [package]
 name = "iota-streams"
 version = "1.0.0"
-authors = ["Vlad Semenov <vlad.semenov@iota.org>", "Dyrell Chapman <dyrell.chapman@iota.org>", "Brord van Wierst <brord@iota.org>"]
+authors = ["Vlad Semenov <vlad.semenov@iota.org>", "Dyrell Chapman <dyrell.chapman@iota.org>", "Brord van Wierst <brord@iota.org>", "Arnau Orriols <arnau.orriols@iota.org"]
```
kwek20, line 6: Welcome!

iota-streams-app/Cargo.toml
```diff
@@ -41,7 +41,7 @@ async-trait = { version = "0.1", optional = true }
 atomic_refcell = { version = "0.1.6", optional = true }
 
 # Dependencies for "client" feature
-iota-client = { git = "https://github.com/iotaledger/iota.rs", branch = "dev", default-features = false, optional = true }
+iota-client = { git = "https://github.com/iotaledger/iota.rs", rev  = "283af58b6f9a20b3698e5128e3e651b2a4dce861", default-features = false, optional = true }
```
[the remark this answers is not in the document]
arnauorriols: wasm is broken in iota.rs, didn't want to spend time bisecting their git log. This commit is the one that was pinned in the `Cargo.lock`, therefore it was the one being used right until we removed the `Cargo.lock`, therefore certain to work with wasm.

bindings/c/src/api/mod.rs
```diff
@@ -526,44 +521,44 @@ pub extern "C" fn drop_payloads(payloads: PacketPayloads) {
 }
 
 #[no_mangle]
-pub unsafe extern "C" fn drop_str(string: *const c_char) {
+pub unsafe extern "C" fn drop_str(string: *mut c_char) {
```
semenov-vladyslav, line 524: Why `mut` here? The idea was that `drop_str` frees strings returned by bindings/c API which is of type `*const c_char`, ie. immutable strings.

bindings/c/src/api/mod.rs
```diff
@@ -526,44 +521,44 @@ pub extern "C" fn drop_payloads(payloads: PacketPayloads) {
...
-pub unsafe extern "C" fn get_address_index_str(address: *mut Address) -> *mut c_char {
-    address.as_ref().map_or(null_mut(), |addr| {
-        get_hash(addr.appinst.as_ref(), addr.msgid.as_ref()).map_or(null_mut(), |index| {
-            CString::new(index).map_or(null_mut(), |index| index.into_raw())
-        })
+pub unsafe extern "C" fn get_address_index_str(address: *const Address) -> *const c_char {
+    address.as_ref().map_or(null(), |addr| {
+        let index = addr.to_msg_index();
+        let index_hex = format!("{:x}", index);
+        CString::new(index_hex).map_or(null(), |index| index.into_raw())
```
semenov-vladyslav, line 561: Maybe factor this pattern out into a separate function `string_into_raw` and replace other occurrences (if any)?

iota-streams-app-channels/src/api/key_store.rs
```diff
@@ -64,8 +69,8 @@ impl<Info, F: PRP> KeyStore<Info, F> for KeyMap<Info> {
                 Identifier::PskId(_id) => self
                     .psks
                     .get_key_value(id)
-                    .filter(|(_, (x, _))| x.is_some())
-                    .map(|(e, (x, _))| (e, x.unwrap().to_vec())),
+                    .map(|(e, (x, _))| x.map(|xx| (e, xx.to_vec())))
```
semenov-vladyslav, line 72: Maybe `filter_map` would be more suitable?

semenov-vladyslav (approved):

Cargo.toml
arnauorriols: :D

bindings/c/src/api/mod.rs
arnauorriols: You are right, don't know what I was thinking. Then I forgot to change the header anyhow, so the example run fine. Actually, I assumed we used `cbindgen` for that. Any reason why we are manually writing the header instead of generating them with `cbindgen`?

iota-streams-app-channels/src/api/key_store.rs
arnauorriols: I thought about it, yeah. However, it does not really suit the scenario. `Option<T>` does not implement `filter_map()` as an specialized method (like it does with `map()`). Therefore, you need to convert the `Option` to an iterator first, polluting the pipeline with `into_iter()` and `collect()`. However, `Option<T>` only implements [`FromIterator<Item=Option<A>> where T: IntoIterator<A>`](https://doc.rust-lang.org/std/option/enum.Option.html#impl-FromIterator%3COption%3CA%3E%3E), which means you cannot build an `Option` out of an iterator unless you store all the items from the iterator in a collection inside the Option, which is not the case. We should probably introduce a `fold()` instead...

Long story short, `flatten()` is the nicest pattern I could come up with (and it's not uncommon).

bindings/c/src/api/mod.rs
arnauorriols: Nice suggestion. Done!

iota-streams-app-channels/src/lib.rs
```diff
@@ -29,3 +29,12 @@ pub mod message;
 
 /// Author and Subscriber API.
 pub mod api;
+
+/// Get a `Value` given a `Key`
+///
+/// This trait can be implemented to any kind of collection to get an item out of it.
+/// It's meant to be versatile, so it can be implemented for T or &T, and both `Key` and
+/// `Value` can be owned or references as well.
+pub trait Lookup<Key, Value> {
```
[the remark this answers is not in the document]
arnauorriols: With something so "silly" as this it didn't even occur to me to search for an exernal implementation... Why would you consider it? 

The left-pad incident always comes to mind when thinking about these kinds of decisions... :joy: Anyway, in Rust this also manifests as a bad idea because of the non-trivial build time introduced unnecessarily, for example, or unnecessary system-dependant compatibility issues from the new dependency and its own dependencies... just to name a few.

bindings/c/CMakeLists.txt
```diff
@@ -4,8 +4,8 @@ project(iota_streams_c LANGUAGES C)
 include(FetchContent)
 
 option(NO_STD "Enable no_std build, without iota_client" OFF)
-option(SYNC_CLIENT "Enable sync transport via iota_client" OFF)
-option(STATIC "Build static library" ON)
+option(SYNC_CLIENT "Enable sync transport via iota_client" ON)
+option(STATIC "Build static library" OFF)
```
DyrellC, line 8: :pray:

bindings/c/src/api/mod.rs
```diff
@@ -61,6 +58,24 @@ pub(crate) fn safe_drop_mut_ptr<T>(p: *mut T) {
     }
 }
 
+/// Convert an String-like collection of bytes into a raw pointer to the first byte
+///
+/// The pointer might be [`null`] if the String contains a null byte (which is invalid)
+///
+/// [`null`]: https://doc.rust-lang.org/std/ptr/fn.null.html
+fn _string_into_raw(string: impl Into<Vec<u8>>) -> *const c_char {
```
DyrellC, line 66: A good pair of utility functions :+1:

bindings/c/src/api/mod.rs
```diff
@@ -533,37 +545,37 @@ pub unsafe extern "C" fn drop_str(string: *const c_char) {
...
+    address.as_ref().map_or(null(), |addr| get_msgid_str(&addr.msgid))
 }
 
 #[no_mangle]
-pub unsafe extern "C" fn get_address_index_str(address: *mut Address) -> *mut c_char {
-    address.as_ref().map_or(null_mut(), |addr| {
-        get_hash(addr.appinst.as_ref(), addr.msgid.as_ref()).map_or(null_mut(), |index| {
-            CString::new(index).map_or(null_mut(), |index| index.into_raw())
-        })
+pub unsafe extern "C" fn get_address_index_str(address: *const Address) -> *const c_char {
```
DyrellC, line 574: Yeah this never needed to have mutable pointers, thanks for correcting that

bindings/c/src/api/auth.rs
```diff
@@ -171,8 +171,10 @@ pub unsafe extern "C" fn auth_send_keyload(
             link_to.as_ref().map_or(Err::NullArgument, |link_to| {
                 psk_ids.as_ref().map_or(Err::NullArgument, |psk_ids| {
                     ke_pks.as_ref().map_or(Err::NullArgument, |ke_pks| {
-                        let identifiers: Vec<Identifier> = ke_pks.into_iter().map(|pk| (*pk).into()).collect();
-                        user.send_keyload(link_to, psk_ids, &identifiers.iter().collect())
+                        let pks = ke_pks.into_iter().copied().map(Into::<Identifier>::into);
```
DyrellC, line 174: Keep an eye out here, Vlad's PR changes this behaviour to have it all input as `Identifiers` instead of as separate collections of `Pk`'s and `PskId`'s. Depending on which order we pull these in we'll just have to make sure to fix it there appropriately. Ideally we should be putting through the other one first.

bindings/c/CMakeLists.txt
DyrellC, line 8: :pray:

bindings/c/src/api/auth.rs
DyrellC, line 174: This logic is going to change with Vlad's PR. We'll be using `Identifier's` directly instead of pulling in `Pk's` and `PskId's` respectively, they'll all have to be converted first and be injected as one unified collection.

bindings/c/src/api/mod.rs
DyrellC, line 66: Fan of these utility functions :+1:

bindings/wasm/src/author/authorw.rs
```diff
@@ -162,15 +162,16 @@ impl Author {
 
     #[wasm_bindgen(catch)]
     pub async fn send_keyload(self, link: Address, psk_ids: PskIdsW, sig_pks: PublicKeysW) -> Result<UserResponse> {
-        let pks: Vec<Identifier> = sig_pks.pks.iter().map(|pk| (*pk).into()).collect();
+        let pks = sig_pks.pks.into_iter().map(Into::<Identifier>::into);
```
DyrellC, line 165: Same thing as before here, we'll be pulling in a collection of `Identifier's` directly. This will also change the `sig_pks` and `psk_ids` input types obviously, so we'll need to keep an eye on that and change them appropriately depending on the order of everything being accepted.

iota-streams-app-channels/src/api/key_store.rs
```diff
@@ -15,7 +15,9 @@ use iota_streams_core::{
 use iota_streams_core_edsig::key_exchange::x25519;
 
 pub trait KeyStore<Info, F: PRP>: Default {
-    fn filter(&self, pks: &[&Identifier]) -> Vec<(&Identifier, Vec<u8>)>;
+    fn filter<'a, I>(&self, pks: I) -> Vec<(&Identifier, Vec<u8>)>
```
DyrellC, line 18: We should rename this from `pks` to `ids` or something, considering it can be either a pk or pskid (and in the future DID). It'll also match the actual implementation below that way too :sweat_smile:

iota-streams-app-channels/src/api/key_store.rs
DyrellC: Yeah I originally tried to use a `filter_map` and ran into that exact scenario. Didn't consider a `fold()` though, that seems like a reasonable alternative.

iota-streams-app-channels/src/api/tangle/author.rs
```diff
@@ -171,15 +171,12 @@ impl<Trans: Transport + Clone> Author<Trans> {
...
-    ///  * `ke_pks`  - Vector of Public Keys to be included in message
-    pub fn send_keyload(
-        &mut self,
-        link_to: &Address,
-        psk_ids: &PskIds,
-        ke_pks: &Vec<&Identifier>,
-    ) -> Result<(Address, Option<Address>)> {
-        self.user.send_keyload(link_to, psk_ids, ke_pks)
+    ///  * `keys`  - Iterable of [`Identifier`] to be included in message
+    pub fn send_keyload<'a, I>(&mut self, link_to: &Address, keys: I) -> Result<(Address, Option<Address>)>
```
DyrellC, line 175: Oh good you've already gone ahead and done this change here, we just need to update the bindings to reflect this same approach.

iota-streams-app-channels/src/api/tangle/user.rs
```diff
@@ -554,10 +569,13 @@ impl<Trans: Transport + Clone> User<Trans> {
         ref_link: &MsgId,
         info: MsgInfo,
     ) -> Result<(Address, Option<Address>)> {
-        let seq = self.user.wrap_sequence(ref_link)?;
```
DyrellC, line 557: There was originally a reason why this was ordered this way but I can't recall the exact reason off the top of my head. If none of the tests failed along the way then whatever it was has likely been rectified already, but I just want to leave a note here noting that there was a purpose behind that at some point. In the meantime I'm content (even elated) with the reordering.

iota-streams-app-channels/src/api/user.rs
```diff
@@ -58,66 +57,43 @@ use crate::{
...
-}
-
-impl<F: PRP, Link: HasLink + fmt::Debug> fmt::Debug for WrapStateSequence<F, Link>
+/// Sequence wrapping object
+///
+/// When using multibranch mode, this wrapping object contains the (wrapped) sequence message ([`WrappedMessage`]) to be
+/// sent and the [`Cursor`] of the user sending it.
+///
+/// When using single-branch mode, only the [`Cursor`] is needed, and no sequence message is sent.
+pub enum WrappedSequence<F, Link>
```
DyrellC, line 73: I think this is a good addition, it'll make single depth more succinct to implement as well.

iota-streams-app-channels/src/api/user.rs
```diff
@@ -1311,3 +1243,34 @@ where
         Ok(user)
     }
 }
+
+impl<'a, F, Link, LG, LS, Keys> Lookup<&Identifier, psk::Psk> for &'a User<F, Link, LG, LS, Keys>
```
DyrellC, line 1247: :pray:

iota-streams-app-channels/src/message/keyload.rs
```diff
@@ -107,23 +112,23 @@ where
...
-                                    .absorb(External(<&NBytes<psk::PskSize>>::from(<&[u8]>::from(&store_id))))?
-                                    .commit()?
-                                    .mask(&self.key),
-                                Identifier::EdPubKey(_pk) => ctx.x25519(
-                                    &x25519::PublicKey::from(<[u8; 32]>::try_from(store_id.as_ref())?),
-                                    &self.key,
-                                ),
-                            }
-                        })
+                ctx.absorb(repeated_keys)?.repeated(&self.keys, |ctx, (id, store_id)| {
```
DyrellC, line 115: Logic here is about to change https://github.com/iotaledger/streams/pull/125/files#diff-2a2e4f5ff0b3d34ef457c25d06f5fc2eaf1a4d0bc6feff728c56f5452e433af5R110 just so you know

iota-streams-app/Cargo.toml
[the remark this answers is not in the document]
DyrellC: It also works to use `branch = "wasm"` but we definitely are going to want to pressure that fix to be a part of dev sooner than later. And then pick a revision we feel comfortable with going forward.

iota-streams-app/src/transport/tangle/mod.rs
```diff
@@ -138,30 +141,95 @@ impl<F> TangleMessage<F> {
 }
 
 /// Tangle representation of a Message Link
-#[derive(Clone)]
+///
```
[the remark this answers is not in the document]
DyrellC: Seconded

iota-streams-app/src/transport/tangle/mod.rs
```diff
@@ -138,30 +141,95 @@ impl<F> TangleMessage<F> {
...
-
-        let msgid =
-            MsgId::from_str(msgid_str).map_err(|e| wrapped_err!(BadHexFormat(appinst_str.into()), WrappedError(e)))?;
+    pub fn new(appinst: AppInst, msgid: MsgId) -> Self {
+        Self { appinst, msgid }
+    }
 
-        Ok(TangleAddress { appinst, msgid })
+    /// Hash the content of the TangleAddress using `Blake2b256`
+    pub fn to_blake2b(&self) -> NBytes<U32> {
```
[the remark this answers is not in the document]
DyrellC: Since we have the `to_msg_index` function now I would agree

iota-streams-app/src/transport/tangle/mod.rs
```diff
@@ -138,30 +141,95 @@ impl<F> TangleMessage<F> {
...
-    pub fn to_string(&self) -> String {
-        let mut address = String::new();
-        address.push_str(&self.appinst.to_string());
-        address.push(':');
-        address.push_str(&self.msgid.to_string());
-        address
+    /// Generate the hash used to index the [`TangleMessage`] published in this address
+    ///
+    /// Currently this hash is computed with [`to_blake2b`].
+    pub fn to_msg_index(&self) -> NBytes<U32> {
```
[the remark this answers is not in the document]
DyrellC: Seems easy enough to just add in now

DyrellC (commented):
Looking really good, couple of smaller comments, adding on to a couple statements Brord made. Depending on how Vlad's PR is looking we should be able to get this pulled in pretty soon here.

iota-streams-app/src/transport/tangle/mod.rs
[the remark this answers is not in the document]
arnauorriols: Working on it!

iota-streams-app/src/transport/tangle/mod.rs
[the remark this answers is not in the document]
arnauorriols: Yeah you are right. I first exposed this method, and later I came up with the more business-oriented `to_msg_index()`, and forgot to hide this one

iota-streams-app/src/transport/tangle/mod.rs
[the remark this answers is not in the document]
arnauorriols: I couldn't introduce it without refactoring the wasm-bindings excessively. So I'm still working on it, but I will create a separate PR for it.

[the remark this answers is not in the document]
arnauorriols: PR opened: #140

iota-streams-app/src/transport/tangle/mod.rs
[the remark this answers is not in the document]
arnauorriols: done

iota-streams-app-channels/src/message/keyload.rs
arnauorriols: Ok! My changes are not logical though, conflict resolution should be straightforward :relieved:

iota-streams-app-channels/src/api/tangle/user.rs
arnauorriols: I have been reviewing past commits but they didn't illuminate me. However, my intuition tells me might be something related with the link store?

iota-streams-app-channels/src/api/key_store.rs
arnauorriols: Done!

bindings/c/src/api/mod.rs
arnauorriols: nice suggestion from @semenov-vladyslav

iota-streams-app-channels/src/api/user.rs
arnauorriols: I was wondering where I should place this little thing. Any alternative suggestion?

bindings/c/src/api/mod.rs
arnauorriols: Yeah... actually the header declarations were already `*const` instead of `*mut`, which makes me wonder... why aren't we using `cbindgen`?

DyrellC: I believe we first tried it out with `cbindgen` and ran into some issues, but those issues might have been related to the ternary processes that we've now removed. Haven't tried using it since Oct/Nov of last year.

iota-streams-app-channels/src/api/user.rs
DyrellC: I think here works for now, not sure where else would really be appropriate

iota-streams-app-channels/src/api/tangle/user.rs
DyrellC: Something along those lines I'm guessing. Anyways it's not of any consequence now so we'll use the changed order

iota-streams-app/Cargo.toml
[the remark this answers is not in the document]
DyrellC: If you wouln't mind updating it to `branch = wasm` for this PR for now, and then we'll update it back once the `iota.rs` wasm bindings pr is merged.

DyrellC (approved):
Looks good, just switch the branch for `iota.rs` and we can merge.

DyrellC:
There is also the conflict in the `wasm/getting_started.md`

arnauorriols:
> There is also the conflict in the `wasm/getting_started.md`

changed iota-client to wasm branch and rebased onto `develop`.

DyrellC:
Lgtm we should be able to merge now

PR #135, opened 2021-08-05
WIP: Codebase review

arnauorriols:
:warning: **WIP**! Not ready for review yet. Comments and opinions are not final and still subject to change as the review progresses

# Interpretation Notes
This review includes many opinions and proposals, some of which might build on top of some others. What this means is that some comments might only make sense in the assumption of hypothetical premises set by other comments. This interpretation notes aim to clarify some of these dependencies:
- All comments build on top of the organizational premise set in [the comment regarding crates organization](https://github.com/iotaledger/streams/pull/135#discussion_r683254112)


src/lib.rs
```diff
@@ -22,6 +22,8 @@
 //! - `iota_streams_ddml`: Modules for working with
 //! the IOTA data description language called DDML, in which all Streams messages are encoded
 
+// CORRECTNESS: Channel is not the high-level API of Streams
```
arnauorriols, line 25: Both the Streams crate module docs and the crate organization suggests that channels is the high-level API of Streams, and core is the low-level. However, channels and <core> have orthogonal goals: <core> means to provide a framework for building cryptographically secure messaging protocols, while channels is one such secure messaging protocol that provides "an organizational tool for structuring and navigating secure data through the Tangle". Kind of like saying "serde-json is the high-level API of serde". This is why from a technical point of view the answer to "what is Streams" is ambiguous, because right now it's multiple things. This is specially relevant since corporate-wide, Streams is being marketed as what channels is, not the cryptographic messaging framework. These naming contradictions lead to confusion and misunderstandings and ultimately, frustration.

Currently, I read this repo as:

core[edsign,keccak]: sponge-based cryptographic engine for signing and encrypting data
DDML: EDSL for en/decoding data into/from byte streams declaratively using "core"
which in itself forms what I understand as "A cryptographic framework for building secure messaging applications protocols" and then:

Channels: "an organizational tool for structuring and navigating secure data through the Tangle. Streams organizes data by ordering it in a uniform and interoperable structure." (from official Streams website)
Besides docs, which are easily rewritten, the important issue here is the crate structure. There should be an univocal client => library public relationship between separated crates. For that, I propose to separate the current unique public crate "Streams" into 2:

spongos: includes core and DDML
iota-streams: includes app and app-channels. I don't have a definitive opinion yet, but I have my doubts about the reusability of app and it's justification as a separate abstraction.

GLOSSARY.md
```diff
@@ -0,0 +1,106 @@
...
+try to redesign the system into a simpler architecture. 
+
+This glossary is organized by contexts. Contexts provide the boundaries within which definitions must hold. Terms might have different
+definitions across contexts, although this should be avoided when possible to avoid complexity. 
+
+# Definitions
+## Channels
+### Author
+||
+|-|
```
kwek20, line 19: Here you still use a table/box, and below theres use of "> explanation"

arnauorriols: I've finally settled for neither, I think e get the best render with italics

.rustfmt.toml
```diff
@@ -2,6 +2,7 @@ unstable_features = true
 imports_layout = "Vertical"
 imports_granularity = "Crate"
 comment_width = 120
+# STYLE&CONVENTION: formatting influences content
```
arnauorriols, line 5: Syntax and layout enforces an influence over the code written. This influence can be specially negative in verbose languages, leading the programmer to take decisions focusing in fighting back the verbosity instead of more valuable factors.

These 2 rustfmt customizations might be causing an extra verbosity burden to the code of this project:

imports_layout = "Vertical": Rust code tends to have many imports, and stacking all of them vertically is a very inefficient way to lay them out, leading programmers to use glob imports (*) to circumvent that (which should really, really not be used except for very limited exceptions)
comment_width = 120: while code width = 80 is (IMHO) overly constrained and we should acknowledge that everybody uses wider screens nowadays, the with of the comments suffers from one very specific problem: code snippets get formatted to 120 characters as well, but cargo-doc renders those snippets in a code-block that's 80chars wide, therefore forcing the appearance of a very annoying horizontal scroll on most snippets.

src/lib.rs
arnauorriols: @DyrellC comment amended after our conversation. Let me know what you think! Thanks :ok_hand:

semenov-vladyslav: Great idea! core-edsig and core-keccak are now moved to core and are just reexports from crypto.rs. Merging ddml and core totally makes sense, I don't think spongos is going to be used by itself anywhere else. Spongos (well, sponge mod mostly) can actually be moved into a separate crate. The proposed `streams` crate which is currently named `app` would be pretty thin (identifier, key_store, hdf, pcf), it's not a bad thing.

> There should be multiple end-user crates

Totally agree, currently we only have one - app-channels.

> an univocal client => library public relationship

Not sure what that means.

PR #138, opened 2021-08-14
Cleanup of conditional dependencies to support a wider range of compilation targets

arnauorriols:
# Description
During preparation of [arnauorriols/iota-streams-esp32], several changes have had to be done in Streams to be able to run it in the ESP32 target, with and without `std` feature enabled. This PR includes all such changes. The following is the list of the most relevant:
- Make `wasm-bindgen` feature of `rand` only enabled when `wasm-client` feature of `app-channels` is enabled (it was always enabled)
- Derive `Debug` in a bunch of types, notably `BucketTransport`, `TangleMessage` and `Keccak1600`
- Get rid of `chrono` dependency. I've taken the assumption that the `timestamp` in `TangleMessage` was a leftover
  from pre-chrysalis, so I've removed it completely. Now `TangleMessage<F>` is an alias of `BinaryMessage<F, TangleAddress>`
- Make `futures` dependency pulled only when a `*-client` feature is enabled (it was always enabled)
- Use `StdRng::from_entropy()` instead of `thread_rng` when `target_os == espidf`

## Type of change

- [x] `cargo check --no-default-features && cargo check && cargo check --all-features`
- [x] `cargo clippy --no-default-features && cargo clippy && cargo clippy --all-features`
- [x] run [arnauorriols/iota-streams-esp32] example
- [x] run examples
- [x] run wasm examples
- [x] run c examples
- [ ] run no-std examples

## Change checklist

- [ ] New and existing unit tests pass locally with my changes

[arnauorriols/iota-streams-esp32]: https://github.com/arnauorriols/iota-streams-esp32


sascha1337:
those 2 commits are causing problems with the node.js bindings - 

> Derive Clone, Debug, PartialEq and Eq for MessageContent 
https://github.com/iotaledger/streams/pull/138/commits/016f61c5c704494995e2b30eb080b201589c6eae

> Fix unused warning for prng::rng when no_std in x25519 wrap command
https://github.com/iotaledger/streams/pull/138/commits/3dffdf63136590debaa8a2dc99e8fb4a90dc2638
again the mio that complains ( and its dependency chain if i remember correctly, had the exact same stuff X times )..

Here's some uncompleted fork, that tried to get the lib working with WASI / WASM, but discontinued efforts:
https://github.com/Thomasdezeeuw/mio/commit/2a792afb1bccab27df8366404c135c605ac825f9

This might be interesting too, something with the core cargo toml dependencies + features declaration is doing weird things, especially with tokio / mio, maybe some helpful hint inside here: 

https://github.com/rust-lang/cargo/issues/9176

I remember actually fixing this issue by tweaking around with the dependencies, overwriting / patching some with that [patch] toml syntax, downgrading a few crates, and using the new feature resolver v2 function on the rust nightly channel ..

https://doc.rust-lang.org/nightly/cargo/reference/features.html#feature-resolver-version-2

But with so many different tries and outdated clones of the streams repo, i cannot remember which exact changes fixed that following build error, that keeps hunting us again and again :D @jlvandenhout 

```sh
deso@nuc /tmp/foo (master *) $ cargo build --target wasm32-unknown-unknown
   Compiling mio v0.7.7
error[E0432]: unresolved import `crate::sys::IoSourceState`
  --> /tmp/cargo-home/registry/src/github.com-1ecc6299db9ec823/mio-0.7.7/src/io_source.rs:12:5
   |
12 | use crate::sys::IoSourceState;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^ no `IoSourceState` in `sys`
```

Otherwise, +1 for your no-std efforts, can't wait to spin up my xargo / cargo xtensa llvm esp32 custom target ^^
Hella fun to completely use cargo instead of pio / arduino / idf.py / esptool.py to build, flash and serial monitor code on ESPs

`cargo espflash --chip esp32 --example esp32 --speed 115200 --release  --features="xtensa-lx-rt/lx6,xtensa-lx/lx6,esp32-hal" /dev/tty.usbserial-0001`

Ive been using this nice crate for it, ive got the Qemu installed but i tend to run it directly on the hardware
> cargo install cargo-espflash

arnauorriols:
Hey @sascha1337 sorry for the late reply! We were focused on releasing 1.2.0 and this PR got backburnered. I've sync'ed this branch with develop, and the wasm build works now.

iota-streams-app/Cargo.toml
```diff
@@ -27,15 +27,15 @@ iota-streams-core = { version = "0.1.2", path = "../iota-streams-core", default-
 iota-streams-core-edsig = { version = "0.1.2", path = "../iota-streams-core-edsig", default-features = false }
 iota-streams-ddml = { version = "0.1.2", path = "../iota-streams-ddml", default-features = false }
 
-# anyhow and chrono are kept in sync with versions used in iota-client
+# anyhow is kept in sync with versions used in iota-client
 anyhow = { version = "1.0.26", default-features = false }
-chrono = { version = "0.4.11", default-features = false, optional = true }
```
arnauorriols, line 32: Now the reasons for removing chrono raised to a whole new level: https://github.com/iotaledger/streams/issues/179
https://github.com/iotaledger/streams/issues/181

sascha1337: chrono ... wasm... tokio... the never ending loop :(

arnauorriols:
> About the timestamp removal. Im all up for it, but when we add it back later (I believe in stardust we get one again?) what is the plan? Or will we be timestamp agnostic, and let the user use iota.rs to fetch timestamps?

Not sure about Stardust, but definitely will in Coordicide. What I'm not convinced of is if this timestamp will be relevant from Streams point of view... Will Streams use the timestamp of the TX for any logic? what will the semantics of this timestamp be? will it be relevant for the subscriber of the Stream? 

In any case, in the present this timestamp is not there, and it is not used, and just adds up to the general confusion. That's why I propose to remove it and deal with it once it makes sense

iota-streams-app/Cargo.toml
```diff
@@ -11,14 +11,14 @@ description = "A rust implementation of the IOTA Streams applications"
 [features]
 default = ["std", "client"]
 # Enable `std` feature in dependencies
-std = ["iota-streams-core/std", "iota-streams-core-edsig/std", "iota-streams-ddml/std", "chrono/std", "chrono/clock", "hex/std"]
+std = ["iota-streams-core/std", "iota-streams-core-edsig/std", "iota-streams-ddml/std", "hex/std"]
```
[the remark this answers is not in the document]
arnauorriols: Well, actually, you helped me realize that rand does not need to be a dependency on `iota-streams-app`. As per `rand/std`, in `iota-streams-core` it is already done as you suggest: https://github.com/iotaledger/streams/pull/138/files#diff-3350ed2612e4263878a1a2e5ba8025625b76c0ec06cc80b7211f586991da47f3R14

DyrellC:
> Will Streams use the timestamp of the TX for any logic? what will the semantics of this timestamp be? will it be relevant for the subscriber of the Stream?

I think it's safe to say that as we currently have it implemented it really doesn't provide much benefit. The only way I could see the timestamp being relevant/beneficial is if it was added to the header so that we can sort multiple messages at a given index by this timestamp, which arguably we shouldn't need to do because (I'm not 100% sure on the stardust side specifically but in corrdicide yes) will contain their own timestamping process that we should be able to use since the timestamps will be embedded into the Tangle Message anyways.

DyrellC (approved):
I see no issues with this one, I think we're good to merge it.

PR #139, opened 2021-08-16
Feature: Single Depth Channels

DyrellC:
# Description of change
Add the third implementation type to Channels. This is the `SingleDepth` implementation. It allows a single publisher (the Author) to send messages that, instead of chaining, surround a single anchor message and can be retrieved by a corresponding sequence number. The retrieval method has been added and can be used like so: 
Rust:
```Rust
let msg = subscriber.receive_msg_by_sequence(anchor_link, msg_num)?;
match msg.body {
    MessageContent::SignedPacket { pk, public_payload, masked_payload } => { ... }, 
    MessageContent::TaggedPacket { public_payload, masked_payload } => { ... }, 
    _ => { Err(...) } // You should only be processing tagged and signed packets
}
```
C:
```C
unwrapped_message_t const *message_return = NULL;
err = auth_receive_msg_by_sequence(&message_return, author, anchor_link, msg_num);
if(e) { 
      printf("Error code: %d\n", e); 
} else { 
    packet_payloads_t response = get_payload(message_return);
    printf("Public: '%s'\n", response.public_payload.ptr);
    printf("Masked: '%s'\n", response.masked_payload.ptr);
}
```

WASM:
```JS
let response = await sub.clone().receive_msg_by_sequence(anchor_link, msg_num);
let msg = response.get_message();
let public_payload = msg.get_public_payload();
let masked_payload = msg.get_masked_payload();
```

## Type of change

A new example has been added to the examples section to demonstrate/test behaviour

## Change checklist

### DIFF
--- a/iota-streams-app-channels/src/api/tangle/subscriber.rs
+++ b/iota-streams-app-channels/src/api/tangle/subscriber.rs
@@ -321,8 +321,8 @@ impl<Trans: Transport + Clone> Subscriber<Trans> {
     ///   # Arguments
     ///   * `anchor_link` - Address of the anchor message for the channel
     ///   * `msg_num` - Sequence of sent message (not counting announce or any keyloads)
-    pub fn receive_msg_by_sequence(&mut self, anchor_link: &Address, msg_num: u32) -> Result<UnwrappedMessage> {
-        self.user.receive_msg_by_sequence(anchor_link, msg_num)
+    pub fn receive_msg_by_sequence_number(&mut self, anchor_link: &Address, msg_num: u32) -> Result<UnwrappedMessage> {
+        self.user.receive_msg_by_sequence_number(anchor_link, msg_num)
     }
 }
 
@@ -492,8 +492,8 @@ impl<Trans: Transport + Clone> Subscriber<Trans> {
     ///   # Arguments
     ///   * `anchor_link` - Address of the anchor message for the channel
     ///   * `msg_num` - Sequence of sent message (not counting announce or any keyloads)
-    pub async fn receive_msg_by_sequence(&mut self, anchor_link: &Address, msg_num: u32) -> Result<UnwrappedMessage> {
-        self.user.receive_msg_by_sequence(anchor_link, msg_num).await
+    pub async fn receive_msg_by_sequence_number(&mut self, anchor_link: &Address, msg_num: u32) -> Result<UnwrappedMessage> {
+        self.user.receive_msg_by_sequence_number(anchor_link, msg_num).await
     }
 }
 
--- a/iota-streams-app-channels/src/api/tangle/subscriber.rs
+++ b/iota-streams-app-channels/src/api/tangle/subscriber.rs
@@ -492,7 +492,11 @@ impl<Trans: Transport + Clone> Subscriber<Trans> {
     ///   # Arguments
     ///   * `anchor_link` - Address of the anchor message for the channel
     ///   * `msg_num` - Sequence of sent message (not counting announce or any keyloads)
-    pub async fn receive_msg_by_sequence_number(&mut self, anchor_link: &Address, msg_num: u32) -> Result<UnwrappedMessage> {
+    pub async fn receive_msg_by_sequence_number(
+        &mut self,
+        anchor_link: &Address,
+        msg_num: u32,
+    ) -> Result<UnwrappedMessage> {
         self.user.receive_msg_by_sequence_number(anchor_link, msg_num).await
     }
 }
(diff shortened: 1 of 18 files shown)


kwek20 (approved):

iota-streams-app-channels/src/api/tangle/user.rs
```diff
@@ -513,6 +514,29 @@ impl<Trans: Transport + Clone> User<Trans> {
...
+    ///   # Arguments
+    ///   * `anchor_link` - Address of the anchor message for the channel
+    ///   * `msg_num` - Sequence of sent message (not counting announce or any keyloads)
+    pub fn receive_msg_by_sequence(&mut self, anchor_link: &Address, msg_num: u32) -> Result<UnwrappedMessage> {
+        if !self.is_single_depth() {
+            return err(ChannelNotSingleDepth);
+        }
+        match self.author_public_key() {
+            Some(pk) => {
+                // Subtract 1 from the sequence number so users can start from 1 instead of 0
```
arnauorriols, line 530: Users are programmers themselves. There's a general assumption among programmers that sequences start from 0, right?

If it is 1, as a user I would appreciate it being explicitly documented

iota-streams-app-channels/src/api/tangle/user.rs
```diff
@@ -513,6 +514,29 @@ impl<Trans: Transport + Clone> User<Trans> {
...
+                let seq_no = self.user.fetch_anchor()?.seq_no - 1;
+                let cursor = Cursor::new_at(anchor_link.rel(), 0, msg_num + seq_no);
+                let link = self.user.link_gen.link_from(&(*pk).into(), cursor);
```
arnauorriols, line 533: This operation copies the `pk` every time, to wrap it into an `Identifier`. This also happens in other places as well. Not proposing to change it necessarily now, but: wouldn't it be more interoperable if `LinkGenerator::link_from` took `AsRef<[u8]>` instead of `&Identifier` for the id? Does `LinkGenerator` really have to concern with what is specifically used as identifier?

iota-streams-app-channels/src/api/tangle/subscriber.rs
```diff
@@ -360,6 +383,9 @@ impl<Trans: Transport + Clone> Subscriber<Trans> {
         public_payload: &Bytes,
         masked_payload: &Bytes,
     ) -> Result<(Address, Option<Address>)> {
+        if self.is_single_depth() {
```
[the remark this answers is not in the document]
arnauorriols: +1. It has a significant UX consequence in that those errors should be compile errors and not runtime errors as they are. There are some methods that return `Result` just for this reason, and shouldn't return `Result` at all. I've noted it in my project review notes and at some point I plan to propose a refactor if not done already by then.

iota-streams-app-channels/src/api/user.rs
```diff
@@ -606,6 +613,10 @@ where
         }
         if !self.is_multi_branching() {
             self.store_state_for_all(msg.link.rel().clone(), seq_no.0 as u32 + 1)?;
+            if self.is_single_depth() {
+                println!("Adding seq no {}", seq_no.0);
```
arnauorriols, line 617: Leftover?

arnauorriols (changes requested):
Good to go from my side, except for these 2 points:
- [why only Author](https://github.com/iotaledger/streams/pull/139/files#r697464915)
- [why start from 1](https://github.com/iotaledger/streams/pull/139/files#r699189443)

and [the println! leftover](https://github.com/iotaledger/streams/pull/139/files#r699288598)

iota-streams-app-channels/src/api/tangle/subscriber.rs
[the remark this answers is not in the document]
DyrellC: As discussed yesterday we'll leave this as is for now and refactor the logic at another time

iota-streams-app-channels/src/api/tangle/user.rs
DyrellC: Yes that does make sense, it would be nice to find a means of making `AsRef<[u8]>` for an `Identifier` but I don't think we'll be able to do so without modifying `Identifier` to not be an enum (or at least stick the type as an enum within a struct with a `Vec<u8>` field) otherwise we will always need to create a new byte vector each time we want to use it instead of just referencing internal fields. Something to consider.

I'll make the `link_from()` change for now and we can consider how to approach the `Identifier` in another PR.

iota-streams-app-channels/src/api/tangle/user.rs
DyrellC: It's back to starting from 0 now

arnauorriols (approved):
Good to go!

Issue #118, opened 2021-07-15
Make .clone() and .copy() internal functions


sobolev-alexey:
## Requirements

Make .clone() and .copy() internal functions and invoke them where needed. 

Simplify library usage and update code examples.

`await author.clone().receive_subscribe(subscriptionLink.copy());` => 
`await author.receive_subscribe(subscriptionLink)`

`await author.clone().get_client().get_link_details(keyloadLink.copy());` => 
`await author.get_client().get_link_details(keyloadLink);`

`await subscriber.clone().get_client().get_link_details(lastLink.copy());` => 
`await subscriber.get_client().get_link_details(lastLink);`

## Are you planning to do it yourself in a pull request?

No.

PR #140, opened 2021-08-21
bindings/wasm: expose `Address#toMsgIndex` (and various consequent refactors)

### LINKED ISSUE #123: bindings/wasm: Expose utility methods for obtaining index IDs
## Description

In the WASM bindings we need to expose a utility method to obtain the Level 1 indexation payload IDs so that they can be consulted on the Explorer or their approval milestones tracked. 

## Motivation

Debugging, demoes, applications.

arnauorriols:
# Description of change

The main goal of this PR is to provide a means to obtain the hash used to index the messages in the Tangle. However, to provide this method a refactor is needed: to convert `Address` into a wrapper around `ApiAddress`, replacing the current string-based implementation that relies on de/serialization for converting between the types. The wrapper pattern provides the same UX as the string-based, and it is a more transparent proxy between Rust and wasm/javascript, simplifying the exposure of this and any other method of the original Rust type. 

This refactor in itself does not break backwards compatibility. However, I take the chance of this PR to propose certain backwards compatibility changes that I deem necessary, albeit maybe they should be done at once for all the API and not just partially like this PR implements. 

- [Convert obvious getter methods into `#[wasm_bindgen(getter)]`][new getter]
- Remove the setter of some properties that don't have a natural setter use case (for example `Address#channelAddress` and `Address#msgId`)
- [Use camelCase instead of snake_case in all javascript API][camelCase]
- [Rename `from_string()` to the more conventional `parse()`][parse]
- [Use `wasm_bindgen_futures::future_to_promise`][future to promise] to get rid of `.copy()` and `.clone()` in the public API, as requested by #118 
- For binary data, [the standard method or property getter should return `Uint8Array`, and we should provide an alternative with the explicit suffix `Hex` for getting the binary as an hexadecimal `String`][hex suffix]. For those cases that the binary is the whole type (like `ChannelAddress` or `MsgId`), the method `bytes()` and `hex()` can be provided to get the `Uint8Array` and hexadecimal `String` respectively.

Besides those, this PR includes also non-breaking cleanup:
- [Remove redundant `#[wasm_bindgen(...)]` attributes][redundant attribute]
- [Error-handling of `AppInst::from_str` and `Msgid::from_str`][error handling]
- Implement Copy for `Address`, `ChannelAddress` (aka `AppInst`) and `MsgId`

Finally, this PR proposes [JSDoc as the automatic API reference documentation generator][jsdoc]. It's by large the most common option in the Javascript ecosystem, which gives a better community, better templates, etc. `wasm-bindgen` does a good job documenting the parameters and return values and merging these JSDoc directives with our own comments (although rustwasm/wasm-bindgen#2394). Although not included in the PR yet, I propose to replace the markdown API-reference of the wasm/javascript bindings with the output of JSDoc. To do this, though, we need to move all the documentation currently in the markdown API-reference to their respective type or method. Using the same docstrings we could use `TypeDoc` to generate Typescript documentation as well without much effort. 

This PR will remain in Draft mode until #132 is resolved, as it builds on changes from that PR. 

Resolves #123

## Type of change

- [x] `cargo +nightly fmt`
- [x] `cargo doc`
- [x] `(cd examples && cargo run)`
- ~`cargo test --workspace`~ (tests of DDML are not maintained, should be fixed in a separate PR)
- [x] `cargo test --workspace --doc --all-features`

## Change checklist

    - [x] Improve `UserResponse::from_strings()` pattern
- [x] I have run Clippy with default and with all features and there isn't any warning
- [x] I have made sure documentation builds and links are not broken
    - [x] Use `Address#toMsgIndex` in examples

[new getter]: https://github.com/iotaledger/streams/pull/140/files?file-filters%5B%5D=.json&file-filters%5B%5D=.md&file-filters%5B%5D=.rs#diff-9b262270f458f9fac5fa91f51bda3ebfecb2e492b63894ad9cfc3693bf09ec52R411-R429
[camelCase]: https://github.com/iotaledger/streams/blob/feb560b07ce69a3615e4875717768fabd6fcbf2f/bindings/wasm/src/types/mod.rs#L174-L178
[parse]: https://github.com/iotaledger/streams/blob/feb560b07ce69a3615e4875717768fabd6fcbf2f/bindings/wasm/src/types/mod.rs#L187-L189
[future to promise]: https://github.com/iotaledger/streams/blob/feb560b07ce69a3615e4875717768fabd6fcbf2f/bindings/wasm/src/user/userw.rs#L36-L52
[redundant attribute]: https://github.com/iotaledger/streams/pull/140/files?file-filters%5B%5D=.json&file-filters%5B%5D=.md&file-filters%5B%5D=.rs#diff-0c567540a740774c29a143c1fffac2b25420291497521636e58921c65bec197bL32
[error handling]: https://github.com/iotaledger/streams/pull/140/files?file-filters%5B%5D=.json&file-filters%5B%5D=.md&file-filters%5B%5D=.rs#diff-03b6b2ccd567c006ee1045f8fe3fc4bd729302f7a28ef67ea8a8717d65869a8aL434-L449
[hex suffix]: https://github.com/iotaledger/streams/blob/feb560b07ce69a3615e4875717768fabd6fcbf2f/bindings/wasm/src/types/mod.rs#L149-L167
[jsdoc]: https://github.com/iotaledger/streams/blob/feb560b07ce69a3615e4875717768fabd6fcbf2f/bindings/wasm/package.json#L22


bindings/wasm/package.json
```diff
@@ -17,7 +17,9 @@
     "prepublishOnly": "npm run build",
     "test": "mocha",
     "serve": "webpack-dev-server",
-    "example:nodejs": "node examples/node.js"
+    "example:nodejs": "node examples/node.js",
+    "example:nodets": "ts-node examples/node.ts",
```
DyrellC, line 21: Nice, I was thinking we should set up a script for the `ts` example too

bindings/wasm/package.json
```diff
@@ -38,16 +40,17 @@
     "@types/node-fetch": "^2.5.10",
     "@wasm-tool/wasm-pack-plugin": "^1.1.0",
     "copy-webpack-plugin": "^9.0",
-    "wasm-pack": "^0.9.1",
     "mocha": "^7.1.2",
     "rimraf": "^3.0.0",
     "ts-node": "^9.1.1",
     "typescript": "^4.3.4",
+    "wasm-pack": "^0.9.1",
```
DyrellC, line 47: There is a `0.10.0` available now, not sure if it's super necessary to update it but it probably wouldn't hurt to get ahead of the ball on that one

bindings/wasm/src/types/mod.rs
```diff
@@ -48,8 +50,29 @@ use iota_streams::{
...
+    /// }
+    /// ```
+    fn into_js_result(self) -> core::result::Result<T, JsValue>;
+}
+
+impl<T, E> ResultExt<T, E> for core::result::Result<T, E>
+where
+    E: ToString,
+{
+    fn into_js_result(self) -> Result<T> {
```
DyrellC, line 73: Big fan of this particular change, the simplification of the higher level api internals is wonderful.

DyrellC (approved):
Don't really have much to say on this one. All the changes look good to me, especially the refactor from `to_result` to `into_js_result`. Definitely makes the `authorw` and `subscriberw` files much cleaner. As for the wrapper around `Address` it just feels like an overdue change at this point so it's good to see it completed.

DyrellC:
Regarding the `Docs` conversation, I agree with the `JSDoc` direction for api docs. We just need to check in with the doc teams to make sure they know that's how the lib will be documented from now on so they make sure to include all the relevant information in the combined doc repositories.

bindings/wasm/package.json
arnauorriols: done

bindings/wasm/src/types/mod.rs
arnauorriols: Yep. Extension traits should be used very sparingly, and the ergonomics improvement needs to be really justified. But when it is, there's no deny it's quite nicer.

kwek20 (approved):

PR #145, opened 2021-09-08
Feat/update sub unsub processes

DyrellC:
# Description of change
Addition of utility functions to aid in subscription and unsubscription. This includes reintroducing the `Unsubscribe` message type as well as adding functions to the `Author` allowing them to manually predefine and remove subscribers from a channel. 

## Type of change

- [x] single branch test has been modified to include author side manual subscription and removal of subscriber A, as well as having subscriber B send an unsubscribe message that is read by 
- [x] wasm bindings have been updated along with the same manual sub/unsub from the author in the examples
- [x] c bindings have been updated along with the same manual sub/unsub from the author in the examples
- [x] all other examples continue to work without being impeded
- [x] checked that state was updated on the author end when subscribers removed (currently no functionality to have those users states removed from other subscribers states) 

## Change checklist


bindings/wasm/src/author/authorw.rs
```diff
@@ -332,4 +344,28 @@ impl Author {
             })
             .into_js_result()
     }
+
+    #[wasm_bindgen(catch)]
+    pub fn store_new_subscriber(&self, pk_str: String) -> Result<()> {
+        public_key_from_string(&pk_str).map_or_else(
+            |err| Err(err),
+            |pk| self.author.borrow_mut().store_new_subscriber(pk).into_js_result(),
+        )
```
arnauorriols, line 353: A more idiomatic way to write the same:
```suggestion
        public_key_from_string(&pk_str)
            .and_then(|pk| self.author.borrow_mut().store_new_subscriber(pk))
            .into_js_result()
```

arnauorriols: This applies also to the other similar methods

bindings/wasm/src/author/authorw.rs
```diff
@@ -332,4 +344,28 @@ impl Author {
             })
             .into_js_result()
     }
+
+    #[wasm_bindgen(catch)]
```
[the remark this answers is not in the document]
arnauorriols: this applies to the other methods as well

examples/src/branching/single_branch.rs
```diff
@@ -363,5 +354,13 @@ pub async fn example<T: Transport>(transport: T, channel_impl: ChannelType, seed
...
     println!("Subscriber states matched");
 
+    println!("\nAuthor unsubscribes Subscriber A");
+    author.remove_subscriber(*subscriberA.get_public_key())?;
+
+    println!("\nSubscriber B sending unsubscribe message");
+    let unsub_link = subscriberB.send_unsubscribe(&subscribeB_link).await?;
+    println!("\nAuthor receiving unsubscribe");
+    author.receive_unsubscribe(&unsub_link).await?;
+
```
arnauorriols, line 364: I would add some lines that test that unsubscribe actually works

iota-streams-app-channels/src/api/key_store.rs
```diff
@@ -31,6 +32,7 @@ pub trait KeyStore<Info, F: PRP>: Default {
     fn keys(&self) -> Vec<(&Identifier, Vec<u8>)>;
     fn iter(&self) -> Vec<(&Identifier, &Info)>;
     fn iter_mut(&mut self) -> Vec<(&Identifier, &mut Info)>;
+    fn remove(&mut self, id: &Identifier) -> Result<()>;
```
arnauorriols, line 35: What's the reasoning behind returning a `Result` here? I could imagine it's meant to support external key-stores, but then all methods should return `Result`...

I wonder the same for the other methods of this trait that return `Result` TBH.

iota-streams-app-channels/src/api/user.rs
```diff
@@ -409,14 +409,77 @@ where
...
+        let preparsed = msg.parse_header().await?;
+        let content = self
+            .unwrap_unsubscribe(preparsed)
+            .await?
+            .commit(&mut self.link_store, info)?;
+        self.remove_subscriber(&content.sig_pk)
+    }
+
+    pub fn remove_subscriber(&mut self, pk: &ed25519::PublicKey) -> Result<()> {
+        let id = (*pk).into();
```
arnauorriols, line 475: It's customary to let the user of a method decide if a value needs copying/cloning. In other words, if a function needs ownership, it should ask the parameter by value instead of copying/cloning internally. In this case, as PublicKey is `Copy`, it would actually clean up the syntactic noise of the borrow and the dereference.

iota-streams-app-channels/src/api/user.rs
```diff
@@ -908,6 +971,10 @@ where
         }
     }
 
+    pub fn remove_psk(&mut self, pskid: &PskId) -> Result<()> {
+        self.key_store.remove(&(*pskid).into())
```
arnauorriols, line 975: Idem, just take `PskId` by value

iota-streams-app-channels/src/api/user.rs
arnauorriols: See https://rust-lang.github.io/api-guidelines/flexibility.html#c-caller-control

arnauorriols (changes requested):
Finished the review. Besides some minor comments to improve idiomacy, the major comment I've left is regarding the `ubsubscribe_key` in [`unsubscribe.rs`](https://github.com/iotaledger/streams/pull/145#discussion_r708531819)

bindings/wasm/src/author/authorw.rs
```diff
@@ -12,7 +12,10 @@ use js_sys::Array;
 
 use core::cell::RefCell;
 
-use iota_streams::app::identifier::Identifier;
+use iota_streams::{
+    app::identifier::Identifier,
+    core::psk::pskid_from_hex_str,
```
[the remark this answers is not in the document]
DyrellC: They should indeed. Updating now

iota-streams-app-channels/src/api/user.rs
```diff
@@ -409,14 +409,77 @@ where
             .await?
             .commit(&mut self.link_store, info)?;
         // TODO: trust content.subscriber_sig_pk
+        // Unwrapped unsubscribe_key is not used explicitly.
```
arnauorriols, line 412: What's the deal with the unsubscribe_key then? Why aren't we using as ["the password to unsubscribe from the channel"](https://github.com/iotaledger/streams/blob/develop/iota-streams-app-channels/src/message/subscribe.rs#L30)?

iota-streams-app-channels/src/message/unsubscribe.rs
```diff
@@ -4,102 +4,112 @@
...
-        ctx: &'c mut wrap::Context<TW, F, OS>,
-    ) -> Result<&'c mut wrap::Context<TW, F, OS>> {
-        let mac = Mac(spongos::Spongos::<TW, F>::MAC_SIZE);
-        ctx.join(store, self.link)?.commit()?.squeeze(&mac)?;
+        ctx: &'c mut wrap::Context<F, OS>,
+    ) -> Result<&'c mut wrap::Context<F, OS>> {
+        ctx.join(store, self.link)?
+            .absorb(&self.sig_kp.public)?
+            .commit()?
+            .ed25519(self.sig_kp, HashSig)?;
```
arnauorriols, line 84: The public key of the subscribers is masked in the `Subscribe` message, but it becomes shared to the other subscribers in the `Keyload` messages. The `unsubscribe_key` is the only key that remains private to the subscriber and the author, therefore if we don't use it at all in here, what's stopping any other user of the channel to forge an `Unsubscribe` message on behalf of another one?

iota-streams-app-channels/src/api/user.rs
DyrellC: Solid question, one I have yet to get an answer from Vlad on. Unsubscribe was removed a long time ago, back before there was a proper signature pair for subscribers, so any logic related to the unsubscribe key used to be based off the spongos state of the unsubscribe message, and verification using a MAC, but now that we have proper signature scheme there's a bit of a toss up on requirement here. 

I don't have a problem with storing the password and using it as access, but the question becomes:  is it necessary to do so if the unsubscribe messages are being signed by the subscriber in the first place? Does the unsubscribe key serve a genuine purpose anymore now that the keypairings are actually being used for verification purposes? 

If not then I think it's a reasonable thing to remove the unsubscribe key altogether, but if it's considered a valuable thing to have, then we need to store that locally, and a password needs to be mapped on the Author end to do a match comparison for manual unsubscription, which may complicate things unnecessarily. I'm open for discussion on the topic though, for now I was more focused on updating the logic to be compatible again as a message type.

iota-streams-app-channels/src/message/unsubscribe.rs
DyrellC: I think the simple answer to that is that the unsubscribe messages are signed by the subscriber instance, and knowing the public key doesn't mean that a user can fake a signature representing that subscriber themselves. If anything i'm more sceptical of the requirement for an unsubscribe key but didn't want to focus too hard on changing the existing logic/internal mechanisms of subscription/unsubscription too much during this iteration

iota-streams-app-channels/src/api/user.rs
DyrellC: Fair points (for this and the public key comment above), I'll update these.

iota-streams-app-channels/src/api/key_store.rs
DyrellC: For this particular remove function, it is pretty unnecessary, I had more logic going on in there at some point and looks like although the logic was removed, the result wasn't, so i'll remove it for now. 

As for the other  functions though, the main reason is due to the possibility of a user trying to make logic that uses an incorrect `Identifier`, for example if a user tries to `insert_psk` with an `Identifier::EdPubKey` it returns an error to inform the user that that operation can't be completed using the Identifier type they've injected. This separation of `Idenitifier` insertions is necessary (at least for now) due to the difference in cursor storage for each type. It would be good to unify them though, then we could avoid the need altogether.

examples/src/branching/single_branch.rs
DyrellC: Done

bindings/wasm/src/author/authorw.rs
DyrellC: I've updated them

bindings/wasm/src/author/authorw.rs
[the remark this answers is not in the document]
DyrellC: I've updated these as well

iota-streams-app-channels/src/message/unsubscribe.rs
arnauorriols: Of course. I was myself pulling backwards from the assumption that there was a reason for the `unsubscribe_key` to begin with.

iota-streams-app-channels/src/api/user.rs
arnauorriols: Very Reasonable. While we wait for @semenov-vladyslav input on this (for a while at least) before taking action, let's change this comment and put a clear `// TODO: remove unused unsubscribe_key because ...`

arnauorriols (approved):

PR #146, opened 2021-09-09
Support selecting transport in examples with env

arnauorriols:
# Description of change

This small PR introduces the possibility to select the transport used in the main Rust examples of the tool.

- by default, the examples run through the bucket transport
- `TRANSPORT=tangle` environment activates the Tangle transport
- `TRANSPORT=bucket` environment explicitly activates the bucket transport

This trivial change has its own PR because it also fixes some lint warnings and runs `cargo +nightly fmt` with a significant amount of noisy changes. 

## Type of change

- [x] (cd examples && cargo +nightly fmt && cargo check && cargo clippy && cargo run)

## Change checklist

### DIFF
--- a/examples/src/branching/multi_branch.rs
+++ b/examples/src/branching/multi_branch.rs
@@ -8,7 +8,7 @@ use iota_streams::{
             ChannelType,
             Subscriber,
             Transport,
-        }
+        },
     },
     core::{
         prelude::HashMap,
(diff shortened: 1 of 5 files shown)


DyrellC (approved):
I don't see any issue with this, seems like a good idea to me, plus the lint fixes help make the example code a bit more readable so not going to complain about that.

Issue #130, opened 2021-07-30
feature: Support falling back to another node


kwek20:
## Description

A secondary node must be added to the Tangle TransportOptions::SendOptions which we will call if a message is not found

## Motivation

Streams data is not stored permanently on the tangle on regular nodes.
In order to still be able to read older data, we want to be able to fall back to another node(likely a permanode)

## Requirements

Write a list of what you want this feature to do.

1. Separate read and write
2. Optionally first check the write node and then the fallback node

## Open questions (optional)
`fetch_next_msgs` and others will fetch until nothing is found, this will always end up calling the fallback node. Is this desired?

LuisMMMTS:
Did not understand if this issue includes the fact that if a node is unnavailable or not responding a second one can be used. I have observed that sometimes when creating a client there is no error given even if the node to which the client connects is in bad health and found no good way of testing its health while creating the client, therefore implementing in client level the fallback would greatly help with error handling

arnauorriols:
@kwek20 is this issue still relevant? Or should we close it?

PR #149, opened 2021-09-15
Feat/wasm: authentication & perma node

### LINKED ISSUE #129: bindings: Add node authorization
## Description

In order to connect to a private node or permanode, we want authorization to be made to protect the node. This is possible in rust but not yet in the bindings wasm and c

## Motivation

username and password on the Client is needed to have secure post on a node

## Requirements

Write a list of what you want this feature to do.

1. Add authorization to wasw,
2. Add authorization to c
3. Update the docs
4. Add an example for both c and wasm

kwek20:
# Description of change

Adds the iots.rs builder options to the WASM bindings
FIxes #129 and maybe #130 

TODO:
C bindings client builder

Needs https://github.com/iotaledger/iota.rs/pull/706 before wasm works (and update back to iotaledger/iota.rs)

## Type of change

ran nodejs, nodets and web examples

## Change checklist

- [X] I have followed the contribution guidelines for this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes


bindings/wasm/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "iota_streams_wasm"
+name = "streams_wasm"
```
DyrellC, line 2: Are we reverting this back now?

bindings/wasm/src/user/builder.rs
```diff
@@ -0,0 +1,274 @@
+// Copyright 2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+#![allow(clippy::unnecessary_wraps)]
```
DyrellC, line 3: Where's the unnecessary wrap that this is here for? Do we want to be introducing new clippy circumvention right after fixing up a bunch of them?

bindings/wasm/src/user/builder.rs
```diff
@@ -0,0 +1,274 @@
...
+        NodeAuthOptions {
+            jwt: self.jwt.clone(),
+            basic_auth_name: self.basic_auth_name.clone(),
+            basic_auth_password: self.basic_auth_password.clone(),
+        }
+    }
+}
+
+#[wasm_bindgen]
+pub struct ClientBuilder(ClientBuilderRust);
```
DyrellC, line 73: Does it make sense for us to provide this pattern through Wasm and C individually like this but not through rust? Would it make sense to migrate the Client Builder wrapper functions over to `iota-streams-app/src/transport/tangle/` somewhere so that they can be accessed by all 3 implementations instead?

DyrellC (changes requested):
I think it makes more sense to put the client builder options inside the rust side of things and then map the bindings to the wrapper functions like we do the `TransportDetails` that way we can map the functionality into wasm and c without having to rewrite them custom for each. There's also the matter of the renaming of the library outputs. Otherwise I like the way the pattern works, very succinct.

bindings/wasm/Cargo.toml
kwek20: thats not reverting!! its supposed to be like that :P idk why it got reverted earlier.
Else NPM has @iota/iota-streams/wasm
that just seems silly :D

bindings/wasm/src/user/builder.rs
kwek20: Oh no, how did that get in there? D:

bindings/wasm/src/user/builder.rs
kwek20: Huh? it is accessible through rust, you just use iota.rs Client builder. So we dont need to have our own copy

bindings/wasm/Cargo.toml
DyrellC: Maybe the earlier reverting was done during Arnau's cleanup refactor a couple weeks back, that checks out then, you can ignore me :stuck_out_tongue:

bindings/wasm/src/user/builder.rs
DyrellC: :scream:

bindings/wasm/src/user/builder.rs
DyrellC: Ah No you're right, I forgot about `Client::new(options: SendOptions, client: iota_client::Client)` since we default to using `new_from_url(url: &str)` so often. I guess all that's needed then is as you said before: making a similar abstraction wrap through the c side... fun fun.

arnauorriols: Aren't we... reimplementing the wasm-bindings of the iota.rs `ClientBuilder`? Why do we need to do that?

bindings/wasm/src/user/userw.rs
```diff
@@ -23,6 +26,13 @@ use iota_streams::{
 #[derive(Clone)]
 pub struct Client(pub(crate) Rc<RefCell<ApiClient>>);
 
+pub fn new_with_client(iota_client: RustClient, options: SendOptions) -> Client {
```
arnauorriols, line 29: Why not a constructor? as in `Client::new_with_client(...)`, not need to make it a js constructor obviously

bindings/wasm/src/user/builder.rs
```diff
@@ -0,0 +1,274 @@
...
+
+    /// Sets after how many seconds new tips will be requested during PoW
+    #[wasm_bindgen(catch, js_name = "tipsInterval")]
+    pub fn with_tips_interval(mut self, tips_interval: u64) -> ClientBuilder {
+        self.0 = self.0.with_tips_interval(tips_interval);
+        self
+    }
+
+    /// Build the Streams Client instance.
+    #[wasm_bindgen(catch)]
```
arnauorriols, line 270: All those `catch` don't belong here, they are used for imported js functions, not exported rust functions

arnauorriols (changes requested):
First batch of review comments. They are mostly cosmetic, except for the more essential one regarding whether aren't we reimplementing the `ClientBuilder` bindings from iota.rs.

arnauorriols (changes requested):
Adding a couple of comments in builder.rs

bindings/wasm/src/author/authorw.rs
```diff
@@ -55,7 +60,19 @@ impl Author {
         Author { author }
     }
 
-    pub fn from_client(client: Client, seed: String, implementation: ChannelType) -> Author {
+    #[wasm_bindgen(catch, js_name = "fromClientRs")]
+    pub fn from_client_rs(client: RustWasmClient, seed: String, implementation: ChannelType) -> Author {
+        let api_client = ApiClient::new_from_client(client.to_inner());
```
arnauorriols, line 67: as per https://rust-lang.github.io/api-guidelines/naming.html#c-conv: 
`client_wasm::Client::to_inner()` should be `client_wasm::Client::into_inner()`

arnauorriols: using `iota_client::Client::clone(&client.into_inner())` we can avoid polluting `ApiClient` with `Rc`, which is a `wasm` artefact.

bindings/wasm/src/author/authorw.rs
```diff
@@ -55,7 +60,19 @@ impl Author {
         Author { author }
     }
 
-    pub fn from_client(client: Client, seed: String, implementation: ChannelType) -> Author {
+    #[wasm_bindgen(catch, js_name = "fromClientRs")]
+    pub fn from_client_rs(client: RustWasmClient, seed: String, implementation: ChannelType) -> Author {
```
arnauorriols, line 66: I would stick with `fromClient()` constructor, because `fromClientRs()` is quite obscure, specially from a javascript point of view. I cannot think of a better name, and for consistency with `import()` and `get_client()`, it make sense that `Author` API only deals with `StreamsClient`

bindings/wasm/src/types/mod.rs
```diff
@@ -358,6 +359,47 @@ impl From<ChannelType> for ApiChannelType {
     }
 }
 
+#[wasm_bindgen]
+pub enum Api {
```
arnauorriols, line 363: This is unused now

bindings/wasm/src/user/userw.rs
```diff
@@ -21,16 +24,23 @@ use iota_streams::{
 
 #[wasm_bindgen]
 #[derive(Clone)]
-pub struct Client(pub(crate) Rc<RefCell<ApiClient>>);
+pub struct StreamsClient(pub(crate) Rc<RefCell<ApiClient>>);
+
+pub fn new_with_client(iota_client: RustClient, options: SendOptions) -> StreamsClient {
```
arnauorriols, line 29: This should be an static method of `StreamsClient` and accept `RustWasmClient` instead of `RustClient`

iota-streams-app/src/transport/tangle/client.rs
```diff
@@ -158,21 +161,20 @@ pub async fn async_get_link_details(client: &iota_client::Client, link: &TangleA
 /// Stub type for iota_client::Client.  Removed: Copy, Default, Clone
 pub struct Client {
     send_opt: SendOptions,
-    client: iota_client::Client,
+    client: Rc<iota_client::Client>,
```
arnauorriols: Cloning `iota_client::Client` in `wasm` we should be able to avoid adding `Rc` here

arnauorriols (changes requested):
Some comments on the new version that reuses `iota.rs`

bindings/wasm/package.json
```diff
@@ -40,17 +40,18 @@
...
     "typescript": "^4.3.4",
     "wasm-pack": "^0.10.0",
     "webpack": "^5.39.1",
     "webpack-cli": "^4",
-    "webpack-dev-server": "^3.7.1",
-    "jsdoc": "^3.6.7"
+    "webpack-dev-server": "^3.7.1"
   },
   "dependencies": {
+    "@iota/client-wasm": "^0.1.2",
```
arnauorriols, line 54: not needed

bindings/wasm/src/author/authorw.rs
arnauorriols: Pushed in a PR to this branch

bindings/wasm/package.json
arnauorriols: pushed in a PR to this branch

iota-streams-app/src/transport/tangle/client.rs
arnauorriols: pushed in a PR to this branch

bindings/wasm/src/user/userw.rs
arnauorriols: pushed in a PR to this branch

bindings/wasm/src/types/mod.rs
arnauorriols: pushed in a PR to this branch

bindings/wasm/src/user/userw.rs
arnauorriols: pushed in a PR to this branch

arnauorriols (approved):

iota-streams-core/src/sponge/spongos.rs
```diff
@@ -306,7 +306,7 @@ impl<F: PRP> Spongos<F> {
 
     /// Only `inner` part of the state may be serialized.
     /// State should be committed.
-    pub fn to_inner(&self) -> Result<Inner<F>> {
+    pub fn into_inner(&self) -> Result<Inner<F>> {
```
arnauorriols, line 309: you went a bit overboard with this change 😂  the gist is:
- **from non-copy owned to owned: `into_`**. 
  - Conveys that the receiver (`self`) is being consumed
- **from ref to owned: `to_`**
  - Conveys that we are getting owned stuff without taking ownership of the receiver (which is usually implemented expensively cloning data)
- **from ref to ref: `as_`**
  - Conveys that we are just getting a view into some internal data, no expensive conversions are made

bindings/wasm/src/author/authorw.rs
```diff
@@ -18,7 +18,9 @@ use iota_streams::{
     app::{
         futures::executor::block_on,
         transport::{
-            tangle::client::Client as ApiClient,
+            tangle::client::{
+                Client as ApiClient,
+            },
```
arnauorriols, line 23: you need to run `cargo +nightly fmt`

bindings/wasm/Cargo.toml
DyrellC: Just going to add my 2 cents here, doesn't need to change in this PR specifically but I think it makes sense to rename to `@iota/streams` and drop the `-wasm`.

bindings/wasm/src/author/authorw.rs
DyrellC: Probably doesn't hurt to throw one more `clippy` in there too just in case.

iota-streams-core/src/sponge/spongos.rs
DyrellC: That seems like a change for a later time/different PR, probably good to just revert this back for now and we'll do that self consumption thing in it's own PR just to keep the separation of concerns in check :sweat_smile:

DyrellC (approved):
Overall looks good, added a couple small comments, mostly things that we'll deal with outside of this PR. Just need to resolve the conflicts and she looks good to go.

bindings/wasm/Cargo.toml
arnauorriols: ok, let's discuss this further in the context of the release

PR #150, opened 2021-09-18
Various cleanups and niceties

arnauorriols:
# Description of change

This PR is a miscellanea of small changes, cleanups and simplifications in preparation for the next PR. The most important changes included are:
- [Remove `PhantomData<F>` from BinaryBody](https://github.com/iotaledger/streams/commit/66c4d04e5555eec5876cc6d01ad7fb63bb7cc0bb)
- [Rename `panic_if_not!()` to `assert!()`](https://github.com/iotaledger/streams/commit/a385cd00f9b4e18e96ceefd138cd720d41deb328)
- [Improve ergonomics of `Bytes`](https://github.com/arnauorriols/streams/pull/150/commits/cbb950c8c843743cf9d8e1c09eabd992af5a6b47)
- [Improve ergonomics of `Identifier`](https://github.com/arnauorriols/streams/pull/150/commits/42913224dbe23e2a0cedbad712e2947a92220b79)
- [Derive Copy and Hash for `Cursor<Link>`](https://github.com/arnauorriols/streams/pull/150/commits/8d028869e835ec1d3507fb8f03ddf6a64071128b)

This PR depends on #146

## Type of change

- Enhancement (a non-breaking change which cleans up code)

## Change checklist


iota-streams-core-edsig/src/key_exchange/x25519.rs
```diff
@@ -78,7 +78,7 @@ pub fn filter_ke_pks<'a>(allowed_pks: &'a Pks, target_pks: &'a [PublicKey]) -> V
         .collect::<Vec<IPk<'a>>>()
 }
 
-#[cfg(test)]
+#[cfg(feature = "deprecated")]
```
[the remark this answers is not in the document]
arnauorriols: I agree, specially since the whole package is deprecated in favour of crypto.rs.

bindings/c/main.c
```diff
@@ -637,6 +615,29 @@ int main()
   printf("\n");
   if(e) goto cleanup;
 
+  {
+    printf("Resetting subscriber state... ");
```
DyrellC, line 619: Was there any particular reason for moving this down or just a matter of making the ordering of actions prettier?

bindings/wasm/examples/node.js
```diff
@@ -103,6 +103,20 @@ async function main() {
     }
   }
 
+  console.log("\nSub sending unsubscribe message");
```
DyrellC, line 106: Is this move to allow the function to pass due to the removal of the subscription link from store in `reset_state`? Or was that just a happy consequence?

iota-streams-app-channels/src/api/tangle/user.rs
```diff
@@ -441,7 +440,8 @@ impl<Trans: Transport + Clone> User<Trans> {
     /// * `link` - Address of message to act as root of previous message fetching
     pub async fn fetch_prev_msg(&mut self, link: &Address) -> Result<UnwrappedMessage> {
         let msg = self.transport.recv_message(link).await?;
-        let header = msg.binary.parse_header().await?.header;
+        let preparsed: Preparsed = msg.binary.parse_header().await?;
```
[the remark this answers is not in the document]
DyrellC: It's because of the removal of `<DefaultF>` from `BinaryMessage` definition I believe

iota-streams-app-channels/src/api/user.rs
```diff
@@ -1007,17 +1007,17 @@ where
         let (
             id,
             Cursor {
-                link: seq_link,
+                link: prev_link,
```
DyrellC, line 1010: We should be careful renaming `seq_link` with `prev_link` because that may be confusing considering the other definition of `previous_link` used while compiling Headers

iota-streams-app/src/identifier.rs
```diff
@@ -78,6 +84,24 @@ impl From<PskId> for Identifier {
     }
 }
 
+impl AsRef<[u8]> for Identifier {
```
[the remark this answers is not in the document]
DyrellC: Agreed

iota-streams-app/src/message/binary.rs
```diff
@@ -14,58 +14,39 @@ use iota_streams_ddml::{
 };
 
 /// Binary Message body with information of how to parse it.
-pub struct BinaryBody<F> {
+#[derive(Clone, Hash, Default, PartialEq, Eq)]
+pub struct BinaryBody {
     pub bytes: Vec<u8>,
```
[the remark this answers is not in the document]
DyrellC: I'm on board with `BinaryBody(Vec<u8>)`. Seems clean and reasonable enough

iota-streams-app/src/message/binary.rs
```diff
@@ -14,58 +14,39 @@ use iota_streams_ddml::{
...
     }
 }
 
-impl<F, Link> BinaryMessage<F, Link>
-where
-    F: PRP,
-    Link: Clone + AbsorbExternalFallback<F> + HasLink + Debug,
-{
-    pub async fn parse_header(&self) -> Result<PreparsedMessage<'_, F, Link>> {
+impl<Link> BinaryMessage<Link> {
```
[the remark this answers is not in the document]
DyrellC: I like the `AbsLink` approach

iota-streams-core-edsig/src/key_exchange/x25519.rs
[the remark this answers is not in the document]
DyrellC: We'll remove all of `_edsig` once we have `crypto.rs` in there so I wouldn't put any emphasis on it for now

iota-streams-core-edsig/src/signature/ed25519.rs
```diff
@@ -15,15 +15,17 @@ use core::hash::{
 
 pub type IPk<'a> = &'a PublicKey;
 
-#[derive(Copy, Clone, Default, Eq)]
+#[derive(Copy, Clone, Default, Eq, Debug)]
 pub struct PublicKeyWrap(pub PublicKey);
 
+// PartialEq implemented manually because Hash is implemented manually
```
DyrellC, line 21: We'll need to check if we'll still need to have wrappers at all once we have `crypto.rs` and if we do, should we be keeping the `_edsig` crate or just moving the logic into `core`?

iota-streams-core/src/errors/error_handler.rs
```diff
@@ -27,12 +27,18 @@ macro_rules! err {
 }
 
 #[macro_export]
-macro_rules! panic_if_not {
-    ($cond:expr) => {{
+macro_rules! assert{
```
DyrellC, line 30: Yes thank you for correcting my idiocy

DyrellC (commented):
Overall i'm on board with the cleanup of everything, just a couple questions/comments specifically regarding the stuff in `iota-streams-core-edsig` but we'll be addressing those in later PR's i'm sure.

bindings/wasm/examples/node.js
arnauorriols: Precisely. I explain the details in the commit message: https://github.com/iotaledger/streams/pull/150/commits/f3dd32b6e7875fb41cafa4dbc2c82a754eb5b8ee

iota-streams-core/src/errors/error_handler.rs
arnauorriols: not idiocy, no! It probably was actually better to call it `panic_if_not` at the beginning. Idiomacy is not fixed, it evolves together with the familiarity of the users: https://www.thefeedbackloop.xyz/stroustrups-rule-and-layering-over-time/

iota-streams-core-edsig/src/signature/ed25519.rs
arnauorriols: I just put these comments here to remind myself in the future not to try get them derived 😆. Once we have `crypto.rs` I don't think it's necessary to keep the wrappers, no, assuming the types provided by crypto.rs provide everything we now 🙄

arnauorriols: but yeah, if we do need the wrappers at the end of the day, putting them at core would be better, less fragmentation

iota-streams-app-channels/src/api/user.rs
arnauorriols: Isn't it the same link? This is my rationale:
- `gen_next_msg_id()` generates the next msg-id of a given user
- for that, it gets the identifier of the user, and the cursor that points to the last message sent by that user
- this cursor includes the link to the last message sent by that user, and its sequence position. From this cocktail, a new msg-id is generated
- this last message link effectively becomes the previous-link of the link being generated

Having said that, I would equally like calling it `last_link`, what do you think?

bindings/c/main.c
arnauorriols: yeah, same as in wasm examples

iota-streams-core-edsig/src/key_exchange/x25519.rs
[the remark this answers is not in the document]
arnauorriols: As long as it gets out of the way of running the tests, it's good enough for me

iota-streams-app-channels/src/api/tangle/user.rs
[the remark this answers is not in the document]
arnauorriols: Exactly. Before, `F` was a type parameter of `BinaryMessage`, so `parse_header()` already knew what `F` is by the time it's called. Now it needs to be inferred when calling `parse_header()`, or in cases like this where it cannot be inferred, it needs to be explicited. I could be explicit about `DefaultF` specifically with a turbofish, but it seemed more relevant to be explicit about the type of the `Preparsed`

iota-streams-app-channels/src/api/user.rs
kwek20: well any link passssed is used as the prev_link so i like the renaming :) 
last_link is fine by me

kwek20 (approved):

kwek20 (approved):
About the members array in cargo.toml; I believe we need to rename iota-streams-app-channels-example into a sub folder called "examples/no-std" (and move original example to an std folder?)

arnauorriols:
@DyrellC @kwek20 What do we want to do with chrono? General consensus is to migrate to time-rs, but I still think we should just remove the timestamp and get rid of the dependency altogether. And if we need/want to properly introduce a timestamp in the Streams messages, we'll deal with the issue then, which whichever crate is best _then_

If you agree, I can remove just the timestamp and the dependency in this PR so that we fix the builds and remove the vulnerability warning, and finish the cleanup in #138

DyrellC:
@arnauorriols I'm of the same opinion, we don't use the timestamp right now for anything and if it's determined that that will be valid to use going forward we can deal with it then. For now we should just remove the dependency and get this bad boy merged.

kwek20 (approved):
LGTM!

DyrellC:
Beautiful looking green checkmark right there. Let's get the ball rolling!

PR #151, opened 2021-09-18
Implement `Messages` stream

### LINKED ISSUE #122: bindings/wasm: gen_next_msg_ids for subscriber
## Description

add gen_next_msg_ids WASM binding for subscriber 

## Motivation

it is missing :)

Explain why this feature is needed.

## Requirements

Write a list of what you want this feature to do.

1. to walk forward in a single branch chain given a message link we currently need to start form the announce link and get all message. It would be very useful be able to walk forward one message by one.

## Are you planning to do it yourself in a pull request?

No.

### LINKED ISSUE #143: Recreation of subscribers in a multi branch channel
## Bug description
I have an unexpected behavior when writing data to a multi branch channel and using a new subscriber instances with the same seed to fetch the data again. 3 messages will be written. All three messages are found by the author. All messages of SubA are found by SubB and all of SubB found by SubA, but not for recreated subscribers.

Here is the code sandbox:
https://codesandbox.io/s/messages-issue-multi-branch-duedb?file=/src/index.js

I would expect that all the logs are every time the same the script is executed. But for the recreated subscribers A & B (below line 284) the logs are different (each time) when it gets executed. Sometime A & B finds all messages in the channel sometimes less or none or one is finding all and the other none etc.. Can you please tell me if there is a problem in my implementation ? In addition if I use the recreate-subscribers.js script for the created channel of the index.js, it again logs different messages (finds less or finds more) as the recreated subscribers of the index.js.  I would expect the recreated subscribers in each of both scripts to always fetch all three messages in the channel.

## Rust version
https://github.com/iotaledger/streams/commit/2964da41b2f5629998f39d1ffa95b76a7ef4263a

## Steps To reproduce the bug

See codesandbox.

## Expected behaviour
- I would expect that all the logs are every time the same the script is executed.
- I would expect the recreated subscribers in each of both scripts to always fetch all three messages in the channel.

### LINKED ISSUE #171: `fetch_all_next_msgs` not exposed to WASM bindings
## Description

Expose `fetch_all_next_msgs` to WASM bindings

## Motivation

To use such a function from Javascript

## Requirements

Expose the functionality following a similar approach than `fetch_next_msgs`

## Are you planning to do it yourself in a pull request?

No

arnauorriols:
# Description of change

Implement `Messages` stream as the backbone of channel traversal. [Streams] are the async equivalent of [Iterators]. As such, they are the standard abstraction (pending stabilization) to iterate over items that are obtained asynchronously. Being the standard, it ensures maximum interoperability with 3rd party libraries and tooling. Not to mention it's a very well thought API, with lots of versatility :) (see [`Messages` docs]).

Some minor breaking changes have been needed to accommodate this abstraction nicely:
- `fetch_next_msgs()` now returns _all_ next messages, much like `fetch_all_next_msgs()` did before
- `fetch_all_next_msgs()` has been removed
- Both `fetch_next_msgs()` and `sync_state()` now return `Result<T>`, propagating transient errors (network for example) that might happen during fetching and collection

Unlike `fetch_all_next_msgs()` and `sync_state()` previously did, this new traversal abstraction makes sure that the channel messages are traversed in pre-order: parent messages are always yielded before children messages. This is necessary in order to keep the Spongos state consistent across users. Thanks to that, this PR fixes #143. Make sure the issue reported is effectively fixed before merging this PR. This PR also closes #122 and also closes #171

This PR depends on #150.

## Type of change

- [x] `cargo +nightly fmt`
- [x] `cargo doc`
- [x] `(cd examples && cargo +nightly fmt)`
- [x] `(cd examples && cargo check)`
- [x] `(cd examples && touch **/*.rs && cargo clippy)`
- [x] `(cd examples && TRANSPORT=bucket cargo run)`
- [x] `(cd examples && TRANSPORT=tangle cargo run)`
- [x] `cargo test --doc --all-features`

## Change checklist

[Streams]: https://docs.rs/futures/0.3.17/futures/stream/index.html
[Iterators]: https://doc.rust-lang.org/std/iter/trait.Iterator.html
[`Messages` docs]: https://github.com/arnauorriols/streams/blob/a8aa28b4d6d20c097f5a17b57f9c76cfef40c9b6/iota-streams-app-channels/src/api/tangle/messages.rs#L46


arnauorriols:
Pending: document `Author::fetch_next_msg()` and `Subscriber::fetch_next_msg()` and use them in the examples of wasm bindings. Reported by https://github.com/iotaledger/streams/issues/122#issuecomment-924906365

arnauorriols:
Mr. Lemonade Discord user has reported a bug that gets fixed by this PR: https://discord.com/channels/397872799483428865/811704697622364251/893441669138120775

arnauorriols:
Pending: support messages linked to sequence messages instead of original messages (as reported in https://discord.com/channels/397872799483428865/811704697622364251/893441669138120775)

iota-streams-app-channels/src/api/tangle/messages.rs
```diff
@@ -0,0 +1,898 @@
...
+///     .await?;
+///
+/// let messages: Vec<UnwrappedMessage> = subscriber
+///     .messages()
+///     .try_skip_while(|msg| {
+///         future::ok(
+///             msg.body
+///                 .masked_payload()
+///                 .and_then(Bytes::as_str)
+///                 .map(|payload| payload != "branch 2")
```
DyrellC, line 410: I think perhaps as an alternative of this example, instead of using `payload != "branch 2"` we could also/alternately use something like: `msg.prev_link == second_keyload_link`. The reason I say this is because in order for the "branch2" approach to work, we need to specify the payload contents of the message that we're trying to filter by and that seems like more of a user implementation choice, whereas filtering by `Keyload` link is more universal. It also serves to remove the need for the `skip(1)` when collecting. 

We don't specify anywhere in our documentation to start a branch with an additional message stating the branch name, so this example might come off as a little confusing.

iota-streams-app-channels/src/api/tangle/messages.rs
```diff
@@ -0,0 +1,898 @@
...
+/// # };
+/// #
+/// # author_process.await?;
+/// # subscriber_process.await?;
+/// # Ok(())
+/// # })
+/// # }
+/// #
+/// ```
+/// ## Filter the messages of a particular branch
```
DyrellC, line 319: Might want to specify that the following is the verbose method before getting into it rather than after, it's a little cluttered down at the part you specify that `filter_branch` condenses the call, I actually missed it when I was scrolling the first time and was wondering why you had the expanded version here when it was already condensed into the other call.

iota-streams-app-channels/src/api/tangle/messages.rs
```diff
@@ -0,0 +1,898 @@
...
+/// network failure, [`Messages::next()`] will return `Err`. It is strongly suggested that, when suitable, use the
+/// methods in [`futures::TryStreamExt`] to make the error-handling much more ergonomic (with the use of `?`) and
+/// shortcircuit the [`futures::Stream`] on the first error.
+pub struct Messages<'a, Trans>(PinBoxFut<'a, (MessagesState<'a, Trans>, Option<Result<UnwrappedMessage>>)>);
+
+type PinBoxFut<'a, T> = Pin<Box<dyn Future<Output = T> + 'a>>;
+
+pub struct MessagesState<'a, Trans> {
+    user: &'a mut User<Trans>,
+    ids_stack: Vec<(Identifier, Cursor<Address>)>,
```
DyrellC, line 506: Would it make sense to allow for an injection into the `ids_stack` through a method to let a user specify the specific starting point they'd like to begin fetching from? I suppose not due to the way it refills it with the next `gen_next_msg_addresses` call and it then goes right back to the end of state. Just curious if there's a way for us to enable walking forward from a specific point without refetching from scratch to that point and then proceeding with `try_next()` or something like that...

iota-streams-app-channels/src/api/tangle/messages.rs
```diff
@@ -0,0 +1,898 @@
...
+            let r = state.next().await;
+            (state, r)
+        }))
+    }
+
+    pub async fn next(&mut self) -> Option<Result<UnwrappedMessage>> {
+        StreamExt::next(self).await
+    }
+
+    /// Start streaming from a particular message
```
DyrellC, line 635: This example seems a little redundant, it looks identical to the one above describing filtering branches. Maybe if in this case we have the filter look for `msg.link != INSERT_STARTING_LINK_HERE` or something like that, otherwise I don't think it makes sense to have it doubled up like this.

iota-streams-app-channels/src/api/tangle/mod.rs
```diff
@@ -120,23 +118,89 @@ impl MessageContent {
...
         }
     }
 
-    pub fn unreadable() -> Self {
-        Self::Unreadable
+    /// Get the masked payload of the message
+    ///
+    /// If the message is a [MessageContent::TaggedPacket] or [MessageContent::SignedPacket]
+    /// it returns `Some(payload)`, otherwise returns `None`.  
+    pub fn masked_payload(&self) -> Option<&Bytes> {
```
DyrellC, line 197: Big fan of all these functions.

iota-streams-app-channels/src/api/tangle/subscriber.rs
```diff
@@ -131,13 +132,19 @@ impl<Trans> Subscriber<Trans> {
...
+    /// Given the set of users registered as participants of the channel and their current registered
+    /// sequencing position, this method generates a set of new [`Address`] to poll for new messages
+    /// (one for each user, represented by its [`Identifier`]). However, beware that it is not recommended to
+    /// use this method as a means to implement message traversal, as there's no guarantee that the addresses
+    /// returned are the immediately next addresses to be processed. use [`Subscriber::messages()`] instead.
+    ///
+    /// Keep in mind that in multi-branch channels, the link returned corresponds to the next sequence message.
+    ///
+    /// The link is returned in a [`Cursor<Link>`] to carry over its sequencing information
+    pub fn gen_next_msg_addresses(&self) -> Vec<(Identifier, Cursor<Address>)> {
```
DyrellC, line 146: Might be worth considering removing this from Subscriber and Author at this point. Seems like just an opportunity for users to get themselves into trouble.

iota-streams-app-channels/src/api/tangle/user.rs
```diff
@@ -409,28 +420,28 @@ impl<Trans: Transport + Clone> User<Trans> {
...
-                if let Ok(msg) = self.handle_message(msg, true).await {
-                    msgs.push(msg);
-                }
-            }
-        }
-        msgs
+    /// Iteratively fetches all the next messages until internal state has caught up
+    ///
+    /// If succeeded, returns the number of messages advanced.
+    pub async fn sync_state(&mut self) -> Result<usize> {
```
DyrellC, line 433: I like the return of number of messages traversed, there's some nice utility to come from that.

iota-streams-app-channels/src/api/user.rs
```diff
@@ -908,47 +863,57 @@ where
...
-            let msg_id = link_gen.link_from(id, Cursor::new_at(&*prev_link, 0, 1));
-            ids.push((*id, Cursor::new_at(msg_id, 0, 1)));
-        } else {
-            let msg_id = link_gen.link_from(id, Cursor::new_at(&*prev_link, 0, *seq_no));
-            ids.push((*id, Cursor::new_at(msg_id, 0, *seq_no)));
-        }
+    /// Generate the link of a sequence message of a user given the previous link of its referred message
+    ///
+    /// The link is returned in a [`Cursor<Link>`] to carry over its sequencing information
+    pub fn gen_seq_link<I>(&self, id: I, previous_link: &Link::Rel) -> Cursor<Link>
```
DyrellC, line 882: Consolidation is our friend

DyrellC (commented):
Generally I'm a fan, I have some comments and questions regarding the examples and something we may want to consider regarding fetching from midway points along a branch.

iota-streams-app-channels/src/api/tangle/messages.rs
```diff
@@ -0,0 +1,898 @@
...
+/// };
+///
+/// #
+/// # use std::cell::RefCell;
+/// # use std::rc::Rc;
+/// # use iota_streams_app_channels::api::tangle::BucketTransport;
+/// # use iota_streams_core::Result;
+/// #
+/// # fn main() -> Result<()> {
+/// # smol::block_on(async {
```
arnauorriols, line 85: Replace `smol` with an alternative, as it is incompatible with `reqwest`. The issue seems to be a known one: https://github.com/seanmonstar/reqwest/issues/941#issuecomment-647794302

iota-streams-app-channels/src/api/user.rs
arnauorriols: indeed

iota-streams-app-channels/src/api/tangle/user.rs
arnauorriols: Yes! I wasn't sure returning the number of messages was the most useful data to return, but it will definitely have some use. This change was driven by https://rust-lang.github.io/api-guidelines/flexibility.html#c-intermediate

iota-streams-app-channels/src/api/tangle/subscriber.rs
arnauorriols: This will be effective as part of the API overhaul of version 1.0

iota-streams-app-channels/src/api/tangle/messages.rs
arnauorriols: yes, I agree. It resulted like this because my train of thought was 1) write example 2) realize it was too complex 3) extract function. I actually wonder if the reader does care about the verbose version at all... maybe I should remove it entirely.

iota-streams-app-channels/src/api/tangle/messages.rs
arnauorriols: Removed. The link to the example in the main docs is enough

iota-streams-app-channels/src/api/tangle/messages.rs
arnauorriols: This _is_ the documentation, and the examples are a good opportunity to suggest patterns. In this case, using the keyload link is not a useful pattern for subscribers, as they are not expected to keep track of keyloads (and doing so is cumbersome compared to sending a message tagging a branch). In any case, I wouldn't spend much more time iterating this, as this is changing soon anyway.

iota-streams-app-channels/src/api/tangle/messages.rs
```diff
@@ -0,0 +1,898 @@
...
+/// let author_seed = "cryptographically-secure-random-author-seed";
+/// let author_transport = Tangle::new_from_url("https://chrysalis-nodes.iota.org");
+/// #
+/// # let author_transport = test_transport.clone();
+/// #
+/// let mut author = Author::new(author_seed, ChannelType::SingleBranch, author_transport);
+/// let announcement_link = author.send_announce().await?;
+/// let shareable_announcement_link = announcement_link.to_string();
+///
+/// // Process 2...
```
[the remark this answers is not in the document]
arnauorriols: If you like this approach, I will change the rest of the examples as well. Let me know

iota-streams-app-channels/src/api/tangle/messages.rs
arnauorriols: The position from which all this machinery starts fetching is governed by the cursor in the state of the user. Both the stack and the stage are just internal buffers necessary to operate. If I understand you correctly, you are asking for the `seek` capacity. I imagine it will be another method of `Messages` that changes the cursor of the user or something similar

iota-streams-app-channels/src/api/tangle/messages.rs
```diff
@@ -0,0 +1,898 @@
...
+///     Author,
+///     Bytes,
+///     ChannelType,
+///     MessageContent,
+///     Subscriber,
+///     Tangle,
+///     UnwrappedMessage,
+/// };
+///
+/// #
```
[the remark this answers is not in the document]
arnauorriols: Sorry I don't get what you mean 😓

arnauorriols:
Conflicts resolved, PR ready to merge

DyrellC (approved):
Alright looking good from my end. Think we're good to merge now.

PR #152, opened 2021-09-22
Add missing `#[wasm_bindgen(constructor)]` attribute to `PublicKeys::new()`

### LINKED ISSUE #115: bindings/wasm: cannot add a new Public Key to PublicKeys array
## Bug description

On the dev branch:

The Public Key of the Subscriber is available.

```
console.log(subs.clone().get_public_key()); 
```

Then I try to add it to a `PublicKeys` instance to be later used in a send_keyload method. 

```
const publicKeys = new PublicKeys();
publicKeys.add(subs.clone().get_public_key());
```

Error is raised:

```
 [Runtime_Error]: null pointer passed to rust
```

I've tried making a copy of the public key string buffer but no luck either. 

```
const str = Buffer.from(Buffer.from(subs.clone().get_public_key())).toString();
```

arnauorriols:
# Description of change

Add missing `#[wasm_bindgen(constructor)]` attribute to `PublicKeys::new()`.

Besides this small fix, that resolves #115, this commit introduces a proposal
to write unit-tests using wasm-bindgen-test. A regression test has been added
reproducing the bug.

## Type of change

- [x] `cargo +nightly fmt`
- [x] `cargo doc`
- [x] `(cd examples && cargo +nightly fmt)`
- [x] `(cd examples && cargo check)`
- [x] `(cd examples && touch **/*.rs && cargo clippy)`
- [x] `cargo test --doc --all-features`
- [x] `(cd bindings/wasm && npm ci && npm run unittest)`
- [x] `(cd examples && TRANSPORT=bucket cargo run)`
- [x] `(cd examples && TRANSPORT=tangle cargo run)`

## Change checklist


DyrellC (approved):
I don't see anything I don't like here, the unit test structure makes sense to me. I'd prefer we keep as many assertions as possible inside the rust end of things but it's also good to be producing minimal type based js examples like you did with `publicKeysWith`. I'd like to set up a naming convention for these ones but I can't think of anything more suitable atm :thinking: so we can stick with this for now. We can approach that again when we tackle adding the rest of the tests.

kwek20 (approved):
I checked out how iota.rs has its tests. They use mocha and JS only.
It does look a bit cleaner to be honest so im not sure if we should be doing it like this.
The only positive thing here is that we will be testing WASM both directions, so it could be more beneficial in the long run.

Lets keep it for now and see how it will feel after adding more tests

Issue #159, opened 2021-10-01
A Subscriber with a seed equal to the Author cannot fetch next messages in a private channel


jmcanterafonseca-iota:
## Bug description

If the Subscriber of a Channel is the same as the Author i.e. their seeds are just the same when the Author sends key loads for everyone the implementation does not take into account such Subscriber.

The `receive_keyload` method returns `true`. The messages published can be received by the Subscriber with `receive_msg`, but when the Subscriber calls `fetch_next_msgs` an empty array is returned. 

## Rust version

WASM bindings

- Rust version:

## Hardware specification

What hardware are you using?

- Operating system:
- RAM:
- Cores:
- Device:

## Steps To reproduce the bug

```ts
    const options = new SendOptions("https://api.lb-0.h.chrysalis-devnet.iota.cafe", true);
    const seed = make_seed(20);
    const auth = new Author(seed,options.clone(),ChannelType.SingleBranch);

    let response = await auth.clone().send_announce();
    let ann_link = response.link;
    const ann_link_copy = ann_link.copy();
    console.log("announced at: ", ann_link.toString());
    console.log("Announce message index: " + ann_link.toMsgIndexHex());
  
    let sub = new Subscriber(seed, options.clone());
    const announce = await sub.clone().receive_announcement(ann_link_copy.copy());

    const response2 = await sub.clone().send_subscribe(ann_link_copy.copy());
    await auth.clone().receive_subscribe(response2.link.copy());

    const result = await auth.clone().send_keyload_for_everyone(ann_link_copy.copy());
    const key_load_link_copy = result.link.copy();
    
    const keyLoad = await sub.clone().receive_keyload(key_load_link_copy.copy());
    console.log("Keyload received: ", keyLoad);

    const response_msg = await sub.clone().send_signed_packet(
        key_load_link_copy.copy(),Buffer.from(""),Buffer.from("hello"));

    let seed3 = seed;
    let sub3 = new Subscriber(seed3, options.clone());
    const announce3 = await sub3.clone().receive_announcement(ann_link_copy.copy());
    const keyLoad3 = await sub3.clone().receive_keyload(key_load_link_copy.copy());
    console.log("Keyload received by Subscriber 3: ", keyLoad3);

    // !!!!! This should return a non empty array!!!!!
    const messages_3 = await sub3.clone().fetch_next_msgs();
    console.log("Fetch next messages subscriber 3:", messages_3.length);

[...]

PR #165, opened 2021-10-04
Don't allow subscription with existing user key

### LINKED ISSUE #164: Reject subscriptions for existing users
## Bug description
Streams `Author's` should not process subscription requests that use the same `Identifier` as an existing user in the channel (especially not their own).

## Steps To reproduce the bug
1. Create an `Author` and send an `Announce` message 
2. Create a `Subscriber` using the same seed as `Author`
3. Send `Subscribe` message to `Author` and receive it

## Expected behaviour
`Subscriber` should not be able to process through step 3. `Author` should fail to receive message because the `Subscriber` has a public key that matches an existing user (`Author`) 

## Actual behaviour
`Subscriber` is being processed and overwriting the `Author` state in store as a result. 

Relates to #159

DyrellC:
# Description of change
Stop `Author` from accepting `Subscribe` messages that are sent from a user with an already registered `Identifier`. Existing user states should not be overwritten. 

Fixes #164 

## Type of change

- Existing examples and binding examples continue to pass (including `streams-examples` repo tests)
- Ran example from #159 and received an error for trying to subscribe with the same seed as author
- Modified example form #159 removing subscription process completes successfully

## Change checklist

### DIFF
--- a/iota-streams-app-channels/src/api/user.rs
+++ b/iota-streams-app-channels/src/api/user.rs
@@ -417,9 +417,10 @@ where
     pub fn insert_subscriber(&mut self, pk: ed25519::PublicKey) -> Result<()> {
         let ref_link = self.appinst.as_ref().unwrap().rel().clone();
         match !self.key_store.contains(&pk.into()) {
-            true => self.key_store
+            true => self
+                .key_store
                 .insert_cursor(pk.into(), Cursor::new_at(ref_link, 0, SEQ_MESSAGE_NUM)),
-            false => err!(UserAlreadyRegistered(hex::encode(pk.as_bytes())))
+            false => err!(UserAlreadyRegistered(hex::encode(pk.as_bytes()))),
         }
     }
 
--- a/iota-streams-app-channels/src/api/user.rs
+++ b/iota-streams-app-channels/src/api/user.rs
@@ -415,12 +415,12 @@ where
     }
 
     pub fn insert_subscriber(&mut self, pk: ed25519::PublicKey) -> Result<()> {
-        let ref_link = self.appinst.as_ref().unwrap().rel().clone();
-        match !self.key_store.contains(&pk.into()) {
-            true => self
+        match (!self.key_store.contains(&pk.into()), &self.appinst) {
+            (_, None) => err!(UserNotRegistered),
+            (true, Some(ref_link)) => self
                 .key_store
-                .insert_cursor(pk.into(), Cursor::new_at(ref_link, 0, SEQ_MESSAGE_NUM)),
-            false => err!(UserAlreadyRegistered(hex::encode(pk.as_bytes()))),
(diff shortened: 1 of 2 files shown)


arnauorriols:
Also fixes #159

iota-streams-app-channels/src/api/user.rs
```diff
@@ -416,8 +416,12 @@ where
...
     pub fn insert_subscriber(&mut self, pk: ed25519::PublicKey) -> Result<()> {
         let ref_link = self.appinst.as_ref().unwrap().rel().clone();
-        self.key_store
-            .insert_cursor(pk.into(), Cursor::new_at(ref_link, 0, SEQ_MESSAGE_NUM))
+        match !self.key_store.contains(&pk.into()) {
+            true => self
+                .key_store
+                .insert_cursor(pk.into(), Cursor::new_at(ref_link, 0, SEQ_MESSAGE_NUM)),
+            false => err!(UserAlreadyRegistered(hex::encode(pk.as_bytes()))),
+        }
```
arnauorriols, line 424: We should probably guard against using this without registering first, and get rid of that `unwrap`. Here's my suggestion:
```suggestion
        match (!self.key_store.contains(&pk.into()), &self.appinst) {
            (_, None) => err!(UserNotRegistered),
            (true, Some(ref_link)) => self
                .key_store
                .insert_cursor(pk.into(), Cursor::new_at(ref_link.rel().clone(), 0, SEQ_MESSAGE_NUM)),
            (false, Some(_)) => err!(UserAlreadyRegistered(hex::encode(pk.as_bytes()))),
        }
```

arnauorriols (approved):
I've left a message suggesting an improvement, but it's not a blocker.

iota-streams-app-channels/src/api/user.rs
DyrellC: I like it, any excuse to remove an unwrap i'm usually on board with.

iota-streams-core/src/errors/error_messages.rs
```diff
@@ -136,7 +136,7 @@ pub enum Errors {
     //////////
     /// Cannot create a channel, user is already registered to channel {0}
     ChannelCreationFailure(String),
-    /// Cannot unwrap announcement message, already registered to channel {0}
+    /// Cannot unwrap message, user already registered to channel {0}
```
kwek20, line 139: This is technically incorrect! Were not doing any message unwrapping here!

Lets change to: `Cannot register new user; user is already registered to channel {0}`

DyrellC: Both uses are done during handle/unwrap instances, but I still think that's a good change for the message anyways, i'll update it

DyrellC: I'm actually going to add the identifier as well, there are 2 places to use it and they're printing separate things, one prints the `appinst` and the other the `identifier`, so i'm going to have both print in the message to make it more concise.

kwek20 (approved):

PR #168, opened 2021-10-05
Feat: Add announcement_link()

### LINKED ISSUE #148: feature: Regain announcement link after import
## Description

Exporting a streams state stores all required data in a file. This file can be used on another device, or another time.
Allegedly, this has all the data to continue a stream. However, new subscribers will need an announcement link which is not available after import on a new run

## Motivation

Without getting the announcement link, we cannot share to new subscribers

## Requirements
1. add method `announce_link` which returns a Result<Link> if a valid state is found

DyrellC:
# Description of change
Add getter function for a user's `appinst` to allow them to get the `Announce` link (provided the user is registered to a channel).

Fixes: #148 

## Type of change

- Existing examples are updated and continue to run 
- Recovery example has an `export/import` added to demonstrate fix for #148 
- C and Wasm binding examples have been updated and run to demonstrate announcement link extraction 

## Change checklist


bindings/wasm/src/subscriber/subscriberw.rs
```diff
@@ -94,6 +94,16 @@ impl Subscriber {
             .into_js_result()
     }
 
+    #[wasm_bindgen(catch, js_name = "announcementLink")]
+    pub fn announcement_link(&self) -> Result<String> {
+        self.subscriber
+            .borrow_mut()
+            .announcement_link()
+            .map(|addr| addr.to_string())
+            .ok_or("channel not created")
```
kwek20, line 103: this msg is weird right? For the auth it makes sense btu not the sub i think

documentation/docs/libraries/wasm/api_reference.md
```diff
@@ -73,6 +73,13 @@ Return the channel address of the channel instance.
 | --------------- | ------------------- | ------------------------- |
 **Returns:** Channel Address for user generated channel.
 
+#### announcement_link(): string
```
kwek20, line 76: I think we need to update this doc, as we have the renaming of all wasm methods t conform the js standard. 
announcement_link -> announcementLink

kwek20 (changes requested):

iota-streams-app-channels/src/api/tangle/user.rs
```diff
@@ -70,6 +70,11 @@ impl<Trans> User<Trans> {
         self.user.appinst.as_ref().map(|x| &x.appinst)
     }
 
+    /// Fetch the Announcement Link of the channel.
+    pub fn announcement_link(&self) -> Option<&TangleAddress> {
+        self.user.appinst.as_ref()
```
arnauorriols, line 75: I would suggest not doing `as_ref()` here, and return `&Option<TangleAddress>` instead. The user is always able of calling `as_ref()`, but the other way around is not possible

bindings/wasm/src/subscriber/subscriberw.rs
DyrellC: A fair point :sweat_smile:  I'll update that

arnauorriols, line 103: Why return `Result` instead of `Option` in the wasm bindings?

documentation/docs/libraries/wasm/api_reference.md
DyrellC: I'll update this as well but we need to also go about updating the rest of them too (probably in its own pr)

iota-streams-app-channels/src/api/tangle/user.rs
```diff
@@ -70,6 +70,11 @@ impl<Trans> User<Trans> {
         self.user.appinst.as_ref().map(|x| &x.appinst)
     }
 
+    /// Fetch the Announcement Link of the channel.
+    pub fn announcement_link(&self) -> Option<&TangleAddress> {
```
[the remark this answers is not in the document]
DyrellC: Just seemed unnecessary to add an additional layer deeper.

bindings/wasm/src/subscriber/subscriberw.rs
DyrellC: I suppose it makes more sense to do it this way, i'll change that, plus it deals with the improper message issue that Brord mentioned as well. 2 birds.

kwek20: Ah i like this better!

iota-streams-app-channels/src/api/tangle/user.rs
DyrellC: This seems reasonable yeah, I removed the `as_ref()` and have it return the reference of the option instead.

documentation/docs/libraries/wasm/api_reference.md
arnauorriols: I have in my todo list to migrate this api reference to JSDoc. I've already communicated this intent to the technical writers team... let's say they haven't opposed to the idea, so that's a go for me :)

arnauorriols (approved):

kwek20 (approved):

Issue #183, opened 2021-10-26
No more able to publish or fetch messages after calling reset_state


dominic22:
## Bug description

Seems like the reset_state function seems to be broken I am no more able to receive the messages from the beginning of a channel using fetch_next_msgs  and in addition I am not able to write new messages to the channel after reset_state was called.

## Rust version
WASM  bindings:
 "@iota/streams": "^0.1.2",

## Steps To reproduce the bug

See codesandbox:
https://codesandbox.io/s/quizzical-heyrovsky-1rsr1?file=/src/index.js

## Expected behaviour

Author/Subscriber is able to fetch the messages again from the beginning of the channel and is also able to publish news messages.

## Actual behaviour

I am no more able to receive the messages from the beginning of a channel using fetch_next_msgs  and in addition I am not able to write new messages to the channel after reset_state was called.

PR #184, opened 2021-10-26
Fix: Keep Announcement state in LinkStore

DyrellC:
# Description of change
`LinkStore` getting reset to default means that all states are removed including the Announcement message state. This PR makes it so that the announcement message state continues to persist in store after being reset. 

#183 

## Type of change

- After resetting state, user was able to re-fetch messages and continue publishing to branch 

## Change checklist

### DIFF
--- a/iota-streams-app-channels/src/api/user.rs
+++ b/iota-streams-app-channels/src/api/user.rs
@@ -262,7 +262,11 @@ where
                     key_store.insert_cursor(*id, Cursor::new_at(appinst.rel().clone(), 0, 2_u32))?;
                 }
                 self.key_store = key_store;
-                self.link_store = LS::default();
+
+                let mut link_store = LS::default();
+                let ann_state = self.link_store.lookup(appinst.rel())?;
+                link_store.update(appinst.rel(), ann_state.0, ann_state.1)?;
+                self.link_store = link_store;
 
                 self.link_gen.reset(appinst.clone());
                 Ok(())


DyrellC:
There's an additional inconsistency we'll need to resolve as well. From a Subscriber point of view, resetting state may allow them to replay the messages from scratch, but they won't be able to fetch their Subscription message since it was deployed outside of the data tree. So unless they recreate the message (in the same way the Author has to recreate the announcement message to verify it was generated the same way) the Subscriber then cannot unsubscribe from the channel without recreating that subscription message and storing it in state again.

arnauorriols:
> There's an additional inconsistency we'll need to resolve as well. From a Subscriber point of view, resetting state may allow them to replay the messages from scratch, but they won't be able to fetch their Subscription message since it was deployed outside of the data tree. So unless they recreate the message (in the same way the Author has to recreate the announcement message to verify it was generated the same way) the Subscriber then cannot unsubscribe from the channel without recreating that subscription message and storing it in state again.

Precisely. The bindings examples are broken actually for this exact reason. I've fixed them in https://github.com/iotaledger/streams/pull/150/commits/f3dd32b6e7875fb41cafa4dbc2c82a754eb5b8ee

DyrellC:
@arnauorriols  @kwek20 Circling back to this issue, I think that there's still a deeper conversation to be had about what it means to "Reset" and how to best communicate that, as well as how to better utilise the already seen messages without having to retrieve them again, but in the shorter term: I think it makes sense to perhaps retain all known subscribers in store, but reset their cursor store to their initial storage position so retrieval is doable again from that announcement message. We can reword it in docs to reflect that as well, but as an interim solution to the issue, making sure that all publisher states are mapped back to the announcement should suffice for now no?

arnauorriols:
For me we could just revert 655d28cb7cc5f1aeab3ac0e008eaa1bb78329034, as we are going to refactor much of this anyway in 1.0. But yes, as long as we solve #183 for now, I'm ok with any solution.

arnauorriols:
Given that we are preparing a new API that will supersede this method with one better suited for rereading the channel, I vote now to merge this PR as is, fixing the issue for the short term, and be done with it

PR #185, opened 2021-10-28
Update streams/documentation/docs/libraries/wasm/getting_started.md

Galman21:
1) new install method through npm (without the need anymore of clone the streams repo and build it manually)
2) get_link() method doesn't exist anymore, it was replaced by link attribute

### DIFF
--- a/documentation/docs/libraries/wasm/getting_started.md
+++ b/documentation/docs/libraries/wasm/getting_started.md
@@ -25,5 +25,5 @@ let response = await author.clone().send_announce();
 
 let ann_link = response.link;
 
-console.log("Channel Announcement at: ", ann_link.to_string());
+console.log("Channel Announcement at: ", ann_link.toString());
 ```
--- a/documentation/docs/libraries/wasm/getting_started.md
+++ b/documentation/docs/libraries/wasm/getting_started.md
@@ -1,39 +1,31 @@
 # Getting Started
-The WASM bindings allow for you to build a Streams API for web applications and nodejs. 
-The streams instance underlying the bindings is built with the `wasm-client` flag to 
-ensure a compatible client interface using the `iota.rs iota-client` crate. 
-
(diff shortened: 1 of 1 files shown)


documentation/docs/libraries/wasm/getting_started.md
```diff
@@ -1,39 +1,31 @@
...
 
-// Author generated with: (Seed, Options, Multi-branching flag)
-let auth = new streams.Author("Unique Seed Here", options, false);
+// Options include: (node-url, local pow)
+let options = new streams.SendOptions(node, true);
 
-// Response formatting: { link, sequence link, msg }
+let client = await new streams.ClientBuilder().node(node).build();
+
+let auth = streams.Author.fromClient(streams.StreamsClient.fromClient(client), "Unique Seed Here", streams.ChannelType.SingleBranch);
```
arnauorriols, line 23: Considering this is a _Getting Started_ example, I would suggest to use the constructor of `Author`, which instantiates the client under the hood. The `Author.fromClient` alternative constructor is meant to be used for customizing the `Client` with the different options that `ClientBuilder` provides. While very useful and needed of an example, here we need to show the most simple way to create an author. What do you think?

While we are at it, I would also rename the variable to `author` instead of `auth`.

arnauorriols (changes requested):
Thanks for the PR! Much appreciated. I left a couple of comments to be resolved before merging

documentation/docs/libraries/wasm/getting_started.md
arnauorriols: Aaand while we are at it, I would also suggest to use `streams.ChannelType.MultiBranch`. We are starting to think that suggesting to use `SingleBranch` instead of `MultiBranch` brings more trouble than it's worth.

Galman21:
Hi Arnau, i submitted the changes you requested :)

PR #186, opened 2021-10-28
Update streams/documentation/docs/libraries/wasm/examples.md

Galman21:
1) added .copy() every time ann_link or sub_link is consumed
2) get_link() method doesn't exist anymore, it was replaced by link attribute


Galman21 (commented):
Hi Arnau, i noticed that you have merged #185 and #187 into develop, but you have not merged #186 into develop. I'm askying if i have to modify something in #186?

arnauorriols:
Not really, we are just looking for an spare moment to give it a proper review! cc/ @DyrellC @kwek20

PR #206, opened 2022-02-25
Fix web bindings build and example

arnauorriols:
# Description of change
Changes performed:
- removed build/web.js: not clear why we needed to mess with the output of wasm-pack, but unless there's a clear justification, medling with it is not warranted. Specifically, our modifications broke the default import of the library
- moved web example dependencies and files into their own directory (`examples/web`)
- Changed webpack config to use `syncWebAssembly`, otherwise the application throws an error when calling any function of the library.

## Type of change

Describe the tests that you ran to verify your changes.
- `cd bindings/wasm && npm run build:web`
- `cd bindings/wasm && npm run examples:web`

## Change checklist

- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation


bindings/wasm/examples/web/webpack.config.js
```diff
@@ -26,16 +26,11 @@ module.exports = {
...
     }),
   ],
-  // Makes the output less verbose
-  stats: 'minimal',
-  // Removes the asset size warning
-  performance: {
-    hints: false,
-  },
   experiments: {
-    asyncWebAssembly: true
```
kwek20, line 39: asyncWeb is the new async way. Why would we downgrade back to the old version? Unless the bugreport was about this, then it can be discussed

kwek20 (commented):
Im not really a fan of this, although it makes the examples clear from wasm and node.
You need to do `npm i` twice. Once in main and once in examples.

Why did you move it all over there?

I think its expected to run `npm i && npm run script`, and then it works.
In this case, npm i succeeds, but the script fails. Seems off to me..

Also, and unrelated to this pr, the web example doesnt function. It 0only compiles :)

arnauorriols:
> Why did you move it all over there?

The web example was spread all over, it needed to get contained into its own directory within `examples/`. Following the same logic, moving the dependencies of the web example to their own package.json improve the cohesion of both the library package.json and the example package.json. Imagine somebody is interested in knowing how to use Streams from the browser. Now it's much easier to know exactly which boilerplate and dependencies are needed. 

Same should be done with the other examples. This way, we have a clear separation between the boilerplate and dependencies of each example, and the dependencies of the library itself. 

Now that I think of it, I think the example should technically be called `examples/webpack`, instead of `examples/web`. If you agree, I can rename. Ideally we would have another example that makes use of the web build of the library, that should be called `web`

bindings/wasm/examples/web/webpack.config.js
arnauorriols: `wasm-bindgen` is not compatible with `asyncWebAssembly` atm. See 
- https://github.com/rustwasm/wasm-pack/issues/835#issuecomment-772591665
- https://github.com/rustwasm/wasm-bindgen/issues/2343

kwek20:
It works now as expected yea! Im fine with changing the name as well. The web example was jsut a POC made in the early wasm testing stages. We should defenitely have someone with actual web experience make a smoother experience :) Maybe a good start for the "streams explorer"

PR #216, opened 2022-03-26
Prepare for new wiki build system

Dr-Electron:
# Description of change

This is a first draft of the new wiki build system. It should make things a lot easier for devs. We will also provide a v2 of the wiki-cli to make things even easier ;)
I will keep this a draft until it is ready, but now if you want you can already look at what we are creating.
The wiki can be used by just running `yarn`, `yarn start` and navigating to http://localhost:3000/streams/welcome

TO-DOs:
- [x] Test
- [x] Add v2 of Wiki CLI

## Type of change

- [ ] Bug fix (a non-breaking change which fixes an issue)
- [ ] Enhancement (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation Fix


Dr-Electron:
Ok, most of the things should work now :trollface:
You can test the docs locally with `yarn start`
I also added a test build to check for broken links.
Also please inform the Wiki team (on discord) once you plan to merge as we need to update the wiki that day too :)

arnauorriols (commented):
Why is it necessary to vendor yarn?

Dr-Electron:
> Why is it necessary to vendor yarn?

What do you mean with vendor yarn? Why we use it?

jlvandenhout:
[The text of this answer](https://stackoverflow.com/a/66468148) was the status at the time we started working on the new build system (we wanted the same package manager version across all docs projects, which modern Yarn ensured by checking in the exact Yarn version).

But with newer versions of Node.js [it seems it is actually not required anymore to check in the binary](https://yarnpkg.com/getting-started/install#about-global-installs).

I think it is best to leave it like this, to have a consistent workflow across docs projects for now, while we are finishing up everything needed for the Stardust release.

Then @Dr-Electron and I will reasses this, I'll create an issue in the Wiki repo to track it.

jlvandenhout:
Tracking issue is here: iota-community/iota-wiki#585

arnauorriols:
Just to clarify, my concern stems from the fact that once the file is merged, it will remain at the git history "forever" (or until we do some history rewrite voodoo), even if you remove it in future commits. In other words, removing it in the future does not resolve my concern. 

Having said that, I won't dare try to fix the Javascript ecosystem 😅 not today anyway.

arnauorriols (approved):

PR #233, opened 2022-04-26
Codebase-wide overhaul

arnauorriols:


spongos/src/core/prp/mod.rs
```diff
@@ -1,13 +1,15 @@
...
+
 /// Pseudo-random permutation.
 ///
 /// Actually, it may be non-bijective as the inverse transform is not used in sponge construction.
 #[allow(clippy::upper_case_acronyms)]
-pub trait PRP: Sized + Default + Clone {
+pub trait PRP {
     /// Size of the outer state in bytes.
     /// In other words, size of data chunk that PRP can process in one transform.
     type RateSize: ArrayLength<u8>;
```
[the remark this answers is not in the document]
DyrellC: As a further note, if we decided that Keccak will be the likely only `PRP` implementation used... do we even need it to be a trait in the first place? If we don't intend to have any other permutation implementations, then this would be yet another oppurtunity to do away with some pesky generics all the way up. Instead of pinning `F = KeccakF1600` in `preparsed` and `user`, we can just have `Spongos` use `KeccakF1600` as default. Just feelsl ike a layer of abstraction we don't need, and it might aid in us getting rid of `generic_array` if the implementations of it keep getting consolidated

LETS/src/id/did.rs
```diff
@@ -0,0 +1,314 @@
...
+        Self::new(
+            decode_b58(did.method_id())
+                .expect("decoding DID method-id")
+                .try_into()
+                .expect("DID method-id vector should fit into a 32 Byte array"),
+        )
+    }
+
+    pub(crate) fn try_to_did(&self) -> Result<IotaDID> {
+        let did_str = DID_CORE.to_string() + &encode_b58(self);
```
DyrellC, line 87: We still need to figure out how we want to represent DID's with their network. We have this pinned structure to them for now, but going forward (when we add the additional Identity logic in) we'll need to address this. Just leaving the note here so we don't forget.

LETS/src/id/identifier.rs
```diff
@@ -0,0 +1,374 @@
...
+impl From<Psk> for Identifier {
+    fn from(psk: Psk) -> Self {
+        Identifier::PskId(psk.to_pskid())
+    }
+}
+
+#[cfg(feature = "did")]
+impl From<&IotaDID> for Identifier {
+    fn from(did: &IotaDID) -> Self {
+        Identifier::DID(DIDMethodId::from_did_unsafe(did))
```
DyrellC, line 171: I don't like having an unsafe conversion in a `From` implmenetation like this, an untouched panic like this would be hard to debug, and could happen in a number of places.

iota-streams/examples/full-example/branching/multi_branch.rs
```diff
@@ -0,0 +1,486 @@
+// Rust
```
DyrellC, line 1: We call it a multi branch example but we only really spawn a single branch and augment it, either we should rename the example or we should split up some of the logic into 2 or more branches.

iota-streams/src/api/key_store.rs
```diff
@@ -0,0 +1,127 @@
...
+
+#[derive(Clone, PartialEq, Eq)]
+pub(crate) struct KeyStore {
+    cursors: HashMap<Identifier, usize>,
+    keys: HashMap<Identifier, x25519::PublicKey>,
+    psks: HashMap<PskId, Psk>,
+}
+
+impl KeyStore {
+    pub(crate) fn new() -> Self {
```
DyrellC, line 27: We just use default so I think we can just remove this function

iota-streams/src/api/key_store.rs
```diff
@@ -0,0 +1,127 @@
...
+        &self,
+    ) -> impl Iterator<Item = (Identifier, x25519::PublicKey)> + ExactSizeIterator + Clone + '_ {
+        self.keys.iter().map(|(identifier, key)| (*identifier, *key))
+    }
+
+    fn contains_psk(&self, pskid: &PskId) -> bool {
+        self.psks.contains_key(pskid)
+    }
+
+    fn get_psk(&self, pskid: &PskId) -> Option<&Psk> {
```
DyrellC, line 69: We never use this function, they are fetched through the `psks()` method

iota-streams/src/api/key_store.rs
```diff
@@ -0,0 +1,127 @@
...
+        self.keys.insert(id, xkey).is_none()
+    }
+
+    pub(crate) fn keys(
+        &self,
+    ) -> impl Iterator<Item = (Identifier, x25519::PublicKey)> + ExactSizeIterator + Clone + '_ {
+        self.keys.iter().map(|(identifier, key)| (*identifier, *key))
+    }
+
+    fn contains_psk(&self, pskid: &PskId) -> bool {
```
DyrellC, line 65: Never use this function either, plus the psk functions will become irrelevant with the migration of psks out of key_store in the future.

iota-streams/src/api/key_store.rs
```diff
@@ -0,0 +1,127 @@
...
+
+    pub(crate) fn get_cursor(&self, id: &Identifier) -> Option<usize> {
+        self.cursors.get(id).copied()
+    }
+
+    pub(crate) fn insert_cursor(&mut self, id: Identifier, cursor: usize) -> bool {
+        self.cursors.insert(id, cursor).is_none()
+    }
+
+    pub(crate) fn remove_cursor(&mut self, id: &Identifier) -> bool {
```
DyrellC, line 43: We never use this function either, we only use the flat `remove()` call

iota-streams/src/api/message.rs
```diff
@@ -0,0 +1,301 @@
...
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct Message {
+    pub address: Address,
+    pub header: HDF,
+    pub content: MessageContent,
+}
+
+impl Message {
+    pub(crate) fn from_lets_message<Unwrap>(address: Address, lets_message: LetsMessage<Unwrap>) -> Self
```
DyrellC, line 41: Is there a reason this isn't a `impl<Unwrap: Into<MessageContent>> From<LetsMessage<Unwrap>> for Message` ?

iota-streams/src/api/user.rs
```diff
@@ -0,0 +1,1013 @@
...
+
+    /// Address of the stream announcement message
+    ///
+    /// None if channel is not created or user is not subscribed.
+    stream_address: Option<Address>,
+
+    author_identifier: Option<Identifier>,
+
+    /// Users' trusted public keys together with additional sequencing info: (msgid, seq_no).
+    id_store: KeyStore,
```
DyrellC, line 124: Is `id_store` the best term for it? I suppose that'll change when we have BranchStore over top of KeyStore anyways, but not sure I like `id_store` really.

iota-streams/src/api/user.rs
```diff
@@ -0,0 +1,1013 @@
...
+    /// Prepare Announcement message.
+    pub async fn announce(&mut self) -> Result<SendResponse<TSR>> {
+        // Check conditions
+        let stream_address = self
+            .state
+            .stream_address
+            .ok_or_else(|| anyhow!("before sending the announcement one must create the stream first"))?;
+
+        // Update own's cursor
+        let user_cursor = ANN_MESSAGE_NUM;
```
DyrellC, line 262: do we need this to be its own variable like this?

iota-streams/src/api/user.rs
```diff
@@ -0,0 +1,1013 @@
...
+    }
+
+    /// Prepare Subscribe message.
+    pub async fn subscribe(&mut self, link_to: MsgId) -> Result<SendResponse<TSR>> {
+        // Check conditions
+        let stream_address = self
+            .stream_address()
+            .ok_or_else(|| anyhow!("before subscribing one must receive the announcement of a stream first"))?;
+
+        let user_cursor = SUB_MESSAGE_NUM;
```
DyrellC, line 291: Same here, seems unnecessary to assign it to a variable of its own

iota-streams/src/api/user.rs
```diff
@@ -0,0 +1,1013 @@
...
+    pub fn add_psk(&mut self, psk: Psk) -> bool {
+        self.state.id_store.insert_psk(psk.to_pskid(), psk)
+    }
+
+    pub fn remove_psk(&mut self, pskid: PskId) -> bool {
+        self.state.id_store.remove_psk(pskid)
+    }
+
+    /// Create a new stream (without announcing it). User now becomes Author.
+    pub fn create_stream(&mut self, channel_idx: usize) -> Result<()> {
```
DyrellC, line 231: Does it provide any specific benefit for us to separate `create` and `announce` logic like this anymore?

iota-streams/src/api/user.rs
```diff
@@ -0,0 +1,1013 @@
...
+            link_to,
+            // Alas, must collect to release the &self immutable borrow
+            self.subscribers()
+                .map(|s| Permissioned::ReadWrite(s, PermissionDuration::Perpetual))
+                .collect::<Vec<_>>(),
+        )
+        .await
+    }
+
+    pub async fn send_signed_packet<P, M>(
```
DyrellC, line 470: A consideration from our end will be: when sending messages with large payloads, do we want to make it so they have to break the payload down before sending it? (In which case we should probably start checking that the totality of public + masked is less than that total) Or do we want to break the payloads down ourselves? And start utilising the PCF carrying frame numbers more directly to allow for large payloads to be broken down and sent all in a row? (In which case we'll need to start returning `Result<Vec<SendResponse<TSR>>>`)

DyrellC (changes requested):
Still some comments and open questions on my end here, but overall the overhaul is looking great. The offline discussions seem to be moving us all in the right direction here.

spongos/src/core/spongos.rs
```diff
@@ -162,101 +164,158 @@ impl<F: PRP> Spongos<F> {
...
+            cipher = &mut cipher[n..];
             self.update(n);
         }
         Ok(())
     }
 
-    /// Encrypt in-place a trit slice with Spongos object.
-    pub fn encrypt_mut(&mut self, mut xyr: impl AsMut<[u8]>) {
+    /// Encrypt in-place a byte slice with Spongos object.
+    pub(crate) fn encrypt_inplace<T>(&mut self, mut xyr: T)
```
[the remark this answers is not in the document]
arnauorriols: > encrypt/decrypt_xor_inplace: Remove - not really sure what additional benefit these provide

Agree

> squeeze_n: Keep - makes sense to give the ability to choose the output size for the hash

Keep in mind that with `squeeze()` you also chose the size of the hash, with the added benefit that you don't need a vector allocation. Examples:
- https://github.com/arnauorriols/streams/blob/6d40eb41c6db0caad819c6cf2847c3be90c787a9/spongos/src/core/spongos.rs#L334
- https://github.com/arnauorriols/streams/blob/7ceac384f6c926c2b7e87ddbbc378e28645cd3b8/LETS/src/address.rs#L256

If you want a vector for some reason, you can always squeeze an array and convert it into a vector. I vote to remove this method

> hash: Remove - Seems surpurflous honestly 
> lower hash: Also remove, seems redundant considering all it does is call the other hash which calls init().sponge() 

Agree. We implement `Digest` anyway, should be enough to express that you can use `Spongos` for hashing

> encrypt_inplace/encrypt/encrypt_n: Remove/replace encrypt_mut with encrypt - I'm not sure I get the benefit of the inplace functions, and as for encrypt and encrypt_n, they rely on default and the user to present the properly sized cipher, just to wind up using encrypt_mut anyways, so I'd be in favor of removing the unused and having one specific encrypt method
> decrypt_inplace/decrypt/decrypt_n: Remove/replace decrypt_mut with decrypt - Same logic as above 

Agree for the most part except with `encrypt` and `decrypt`. Notice that I just pushed a fix for them, that tights the input type with the output type so that the type system ensures that both are of the same size. With this change, these for me should actually be the canonical methods to encrypt or decrypt data, if only the DDML hadn't this weird way to manipulate the buffer. I actually have in my radar to refactor the `try_advance()` method of the DDML Contexts (yes, I left improvements out of this PR 🤣 ), so that they implement the Write and Read traits, thus manipulating the buffer in a more standard pattern. Once we had that, the proper method would be `Spongos::encrypt()` instead of `Spongos::encrypt_mut()` (idem for decrypt). 

> is_committed: Undecided, there may be a reasonable use for this function so i'm leaning more towards keeping than removing

Me too. TBH, I'm concerned about the fragility of the Spongos and the dependency of its security to the commit operation. I have it in my radar to design some sort of RAII guard pattern to make it less error-prone. For now though, I'd keep this harmless function around.

spongos/src/ddml/commands/sizeof/absorb.rs
```diff
@@ -0,0 +1,186 @@
...
+{
+    fn absorb(&mut self, bytes: Bytes<&T>) -> Result<&mut Self> {
+        let bytes_size = Size::new(bytes.len());
+        self.absorb(bytes_size)?;
+        self.size += bytes.len();
+        Ok(self)
+    }
+}
+
+impl<'a, T> Absorb<&'a Bytes<T>> for Context
```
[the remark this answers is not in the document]
arnauorriols: I agree

spongos/src/ddml/commands/sizeof/absorb.rs
```diff
@@ -0,0 +1,186 @@
...
+where
+    Self: Absorb<External<NBytes<&'a T>>>,
+{
+    fn absorb(&mut self, external: External<&'a NBytes<T>>) -> Result<&mut Self> {
+        self.absorb(External::new(NBytes::new(external.into_inner().inner())))
+    }
+}
+
+// Implement &External<T> for any External<&T> implementation
+impl<'a, T> Absorb<&'a External<T>> for Context
```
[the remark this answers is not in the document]
arnauorriols: Agree

spongos/src/ddml/types/nbytes.rs
```diff
@@ -0,0 +1,93 @@
...
+impl<T> AsMut<[u8]> for NBytes<T>
+where
+    T: AsMut<[u8]>,
+{
+    fn as_mut(&mut self) -> &mut [u8] {
+        self.0.as_mut()
+    }
+}
+
+impl<T, Idx> Index<Idx> for NBytes<T>
```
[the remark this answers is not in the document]
arnauorriols: These are used, I think you are confusing them with `Indexable` and `Addressable` and all that stuff that I've already removed

spongos/src/core/prp/mod.rs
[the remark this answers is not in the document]
arnauorriols: While I advocate for removing unnecessary generics from the application layer (ie iota-streams), I do think that at the spongos level its reasonable to have the PRP abstracted. Particularly considering it's a work already done, I lean towards keeping the abstraction. 

We could remove GenericArray from PRP with a bit of work, if we really feel like it. I am content already though having it constrained  within `PRP` implementations (and `Digest`, but that's out of our reach).

LETS/src/id/identifier.rs
arnauorriols: I agree, although I'm not sure about fixing it now, considering the WIP status of DID integration. Particularly, I still want to see how a DID identifier ends up looking like, as I think it will likely not be this current implementation using just the method ID. 

This is why I intended to make everything more explicit, so that there's the clear indicator that 1) the method-id is the sole piece of the DID currently being used as identifier, and 2) the conversion from did to a 32Byte array is unsafe.

In other words, the problem is that I don't think the solution should be to use `TryFrom`, but actually to make the DID identifier variable in size (and that's a job for another PR).

iota-streams/examples/full-example/branching/multi_branch.rs
arnauorriols: I agree, planning on a dedicated PR(s) for documentation, including these examples. I will rename for now

iota-streams/src/api/key_store.rs
arnauorriols: I think it's slightly more idiomatic to write `KeyStore::new()` than `KeyStore:;default()`, even though they are the same. See https://rust-lang.github.io/api-guidelines/interoperability.html?highlight=default#types-eagerly-implement-common-traits-c-common-traits

> Note that it is common and expected for types to implement both Default and an empty new constructor. new is the constructor convention in Rust, and users expect it to exist, so if it is reasonable for the basic constructor to take no arguments, then it should, even if it is functionally identical to default.

iota-streams/src/api/key_store.rs
arnauorriols: True. I added it for consistency with `get_key()` and `get_cursor()`, but as long as the `KeyStore` type remains private within the crate, there's little point in maintaining it. I'll remove it

iota-streams/src/api/key_store.rs
arnauorriols: 👍

iota-streams/src/api/key_store.rs
arnauorriols: 👍

iota-streams/src/api/message.rs
arnauorriols: Yes. the `Address` parameter.

iota-streams/src/api/user.rs
arnauorriols: LOL no, definitely not the best term. neither is keystore, nor cursorstore, though. At least, the common denominator of all the stuff stored in there are the `Identifiers` of the subscribers, mapped to keys and to cursors. 

Considering the `PSK` will become their own thing, and the keys will go away once we implement key exchange within `Identity`, this will essentially become just the `CursorStore` (with branch-awareness, but still I think it should be called `CursorStore`). I would change the name when these changes actually happen though, as things are always in motion and one can never really now about the future 😄

iota-streams/src/api/user.rs
arnauorriols: I had a TODO comment about merging the two right until the end. However, there's an appeal into keeping the two operations separated, particularly being `announce` an IO operation. Thinking about semantics, `announce()` does not really convey that the channel is being created, merely that it is being announced (pedantic, I know). 

Lately I've been thinking we should also have some sort of RAII guard to ensure at compile time that the operations that rely on having an appaddr indeed have it. However, it is true that unless we have something of the sort, having to call `create_stream()` explicitly is an awful UX. 

I will merge the two for now.

iota-streams/src/api/user.rs
arnauorriols: Good one. In my opinion, it would be a breach of contract if these methods, that from the point of view accept 1 single message, end up returning a vector of responses. So, regardless of the decision, I think we should stick with the current public API. 

This means that if we want to end up implementing the message fragmentation, we should also work out the response defragmentation behind the scenes. Don't really want to get into how we should do that, but now that you mention it, I do think it's very interesting to provide such feature. It's often argued that "you don't want to send big files anyway, streams is not meant for that", but these arguments fail to see the great efficiency in sending big batches of small data points in a single packet. 

Having said all that, as the limitation is transport specific, we shouldn't perform the check at application level, but at the transport implementation. And I guess iota.rs `Client` already has the check at some point, doesn't it?

spongos/src/ddml/types/nbytes.rs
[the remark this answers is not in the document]
DyrellC: Yes I am

iota-streams/src/api/key_store.rs
DyrellC: Ok I can appreciate that

iota-streams/src/api/user.rs
DyrellC: I prefer `create_stream()` as well. It's more self explanatory

DyrellC:
Overall looking good now. Suppose we just need to update the workflows to build properly and I'm content with proceeding.

iota-streams/src/api/user.rs
kwek20: Feels unneeded to me, just to add a bit of "readability" (if that was the goal)

arnauorriols: Nah, just copied from the other methods and changed this value. Although I have to mention I do like to keep the implementation of the different methods as verbatim as possible. I like to make it obvious some code is a copy of another. But it's just as obvious, so nevermind.

I'll change it now

spongos/src/core/prp/mod.rs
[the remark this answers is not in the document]
kwek20: Im also in favour of leaving it abstracted in spongos. GenericArray to me is fine to keep. Dont think itll get prettier or faster with just typenum :D

DyrellC (approved):
Looking good, looks like it's about time to merge

kwek20 (approved):

PR #235, opened 2022-05-02
updated did to latest rev

kwek20:
# Description of change

Fixed iota-crypto build error
see: https://github.com/iotaledger/identity.rs/pull/822

## Type of change

### DIFF
--- a/iota-streams-ddml/Cargo.toml
+++ b/iota-streams-ddml/Cargo.toml
@@ -17,4 +17,4 @@ std = ["rand/std", "iota-streams-core/std"]
 iota-streams-core = { version = "0.1.2", path = "../iota-streams-core", default-features = false }
 rand = { version = "0.8", default-features = false }
 hashbrown = { version = "0.8.2", default-features = false, optional = false, features = ["ahash"] }
-iota-crypto = { version = "0.9.1", default-features = false, features = ["x25519", "rand"] }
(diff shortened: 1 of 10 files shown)


DyrellC (approved):
Looks good to me

iota-streams-ddml/Cargo.toml
```diff
@@ -17,4 +17,4 @@ std = ["rand/std", "iota-streams-core/std"]
 iota-streams-core = { version = "0.1.2", path = "../iota-streams-core", default-features = false }
 rand = { version = "0.8", default-features = false }
 hashbrown = { version = "0.8.2", default-features = false, optional = false, features = ["ahash"] }
-iota-crypto = { version = "0.9.1", default-features = false, features = ["x25519", "rand"] }
\ No newline at end of file
+iota-crypto = { version = "=0.9.1", default-features = false, features = ["x25519", "rand"] }
```
arnauorriols, line 20: pining version is not the right solution here. Luckily this is no longer needed because the incorrectly versioned version of iota-crypto has been yanked. Updating DID is good anyway, but remove the pinning and we can merge this PR

arnauorriols (changes requested):
Now that the pinning is no longer necessary, it should be removed. The update of did is a good thing anyway.

Issue #245, opened 2022-05-26
Implement key-exchange for DID using x25519 verification method


arnauorriols:
# DoD
- Key encryption is performed by `Identifier`, key decryption is performed by `Identity`
- Neither `Subscription`, `Keyload` or `Announcement` messages carry KE keys. These are derived from the `Identifier` of the users

Issue #240, opened 2022-05-26
Separate `KeyStore` from `CursorStore`


arnauorriols:
https://github.com/iotaledger/streams/pull/237#discussion_r881533293

PR #251, opened 2022-05-31
[v2.0] Move exchange keys into User

DyrellC:
# Description of change
Remove exchange key management from `KeyStore` and rename to `CursorStore`. Exchange keys will now be hosted inside the `User` for now (will be removed later once #245 is addressed). 

Addresses: #240 

## Type of change

- examples and tests continue to run with changes
- cargo fmt/clippy/test


arnauorriols (approved):
All good! I would say something about the method naming of `CursorStore`, but I assume they will change again the topic PR

streams/src/api/user.rs
```diff
@@ -854,18 +864,18 @@ impl<'a> ContentWrap<State> for wrap::Context<&'a mut [u8]> {
...
-        let cursors = user_state.id_store.cursors();
+        let cursors = user_state.cursor_store.cursors();
         let amount_cursors = cursors.len();
         self.mask(Size::new(amount_cursors))?;
         for (subscriber, cursor) in cursors {
             self.mask(&subscriber)?.mask(Size::new(cursor))?;
         }
 
-        let keys = user_state.id_store.keys();
+        let keys = user_state.exchange_keys.iter().map(|(id, xkey)| (id, xkey));
```
kwek20, line 874: same for here

kwek20 (changes requested):
lets clean this,  then good to go!

streams/src/api/user.rs
```diff
@@ -816,18 +826,18 @@ impl ContentSizeof<State> for sizeof::Context {
...
```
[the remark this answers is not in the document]
DyrellC: Yes, yes we can

streams/src/api/user.rs
DyrellC: done

Issue #258, opened 2022-06-10
branch announcement message type


arnauorriols:
# Description
Stream announcements communicate the _stream_ author's identifier to the rest of the participants. _Branch_ announcements, instead, communicate (or will need to comunicate) the topic of the new branch. Apart from that, branch announcements must be chained so that only the intended subscribers can create and read it, and more importantly, inherit the permissions to the new branch. 

As a consequence of the requirements stated above, branch announcements cannot be read by anybody and cannot be the message whose Spongos keyload join (because keyloads need to be readable by anybody at any times). Therefore, keyloads should revert to joining the stream announcement spongos, and link to the tip of the branch. 

# DoD
- branches are created sending a message of type `BranchAnnouncement`
- streams are created sending a message of type `StreamAnnouncement` (ie rename `Announcement` to `StreamAnnouncement`)
- `BranchAnnouncement` messages include the topic of the new branch being created
- `BranchAnnouncement` messages are sent to the parent branch from which the new branch is being forked (therefore, the topic in the HDF of the `BranchAnnouncement` message is the parent branch)
- the cursor of the creator of the branch is increased in the parent branch (naturally, as the message is sent there)
- the publishers being tracked in the parent branch at the moment of creation of the branch are copied to the new branch, with the INIT cursor
- `BranchAnnouncement` message are logically linked to the last message sent in the branch, just like any other message
- `Keyload` messages are logically linked to the last message sent in the branch, just like any other message
- however, `Keyload` messages join the `StreamAnnouncement` spongos state instead of the linked message Spongos state

DyrellC:
My initial thoughts are that it makes sense to have a separate announcement message for branches specifically, but considering the nested nature of the branches, I think it makes more sense for `Keyload` messages to link their spongos state to the `BranchAnnoncement` messages rather than the `StreamAnnouncement`. The attachment to `StreamAnnouncement` makes the `Keyload` readable for everyone in the channel, but if we intend to have `BranchAnnouncement` only be visible to those who had permissions to walk up the previous branches, then the same logic should apply to the `Keyload` messages attached above them.

kwek20:
The point is that we dont want to need to send new keyloads for each new branch. By that logic, a keyload need to come before a topic, so it can have a link to the permissions.

DyrellC:
That's what I'm saying though, if you have
```
[...]

PR #262, opened 2022-06-27
[v2.0] Branch Announcement Message

DyrellC:
# Description of change

Adds a new `BranchAnnouncement` type message that conveys linked and new `Topic` details. Separates the announcement logic into `StreamAnnouncement` and `BranchAnnouncement`. 
Addresses #258 

## Type of change

- Examples have been adjusted to check for `BranchAnnouncement` messages while fetching 
- Updated `Messages` test 
- `fmt` and `clippy`


streams/src/api/user.rs
```diff
@@ -590,58 +651,59 @@ where
...
+            .insert_cursor(&prev_topic, identifier.clone(), self.next_cursor(&prev_topic)?);
         self.state
             .cursor_store
             .insert_cursor(&topic, identifier.clone(), INIT_MESSAGE_NUM);
         self.state.spongos_store.insert(address.relative(), spongos);
 
         // Update branch links
-        self.state.cursor_store.set_anchor(topic, address.relative());
-        self.state.cursor_store.set_latest_link(topic, address.relative());
+        self.state.cursor_store.set_anchor(&topic, address.relative());
```
arnauorriols, line 705: Once `Keyload` messages revert to linking to the `Announcement` message instead of the `BranchAnnouncement` messages, the tracking of the anchor will not be needed anymore. Do we want to do do this change now maybe? it's just a revert to an already reviewed code, shouldn't be much hassle, and we will be able to consider this functionality completely done.

arnauorriols (commented):
Looking very good. Just a couple of things that I find missing, and I think we could have them in this same PR so that we can consider this functionality done already.

streams/src/api/user.rs
```diff
@@ -266,20 +258,61 @@ impl<T> User<T> {
...
+        let cursor = preparsed.header().sequence();
+
+        // Insert new branch into store
+        self.state.cursor_store.new_branch(topic.clone());
+        // From the point of view of cursor tracking, the message exists, regardless of the validity or
+        // accessibility to its content. Therefore we must update the cursor of the publisher before
+        // handling the message
+        self.state
+            .cursor_store
+            .insert_cursor(&topic, publisher.clone(), INIT_MESSAGE_NUM);
```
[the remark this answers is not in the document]
DyrellC: Probably reasonable to do that yeah. I'll look at implementing it in this PR.

streams/src/api/user.rs
```diff
@@ -266,20 +258,61 @@ impl<T> User<T> {
...
+            }
+        };
+        let branch_announcement = branch_announcement::Unwrap::new(&mut linked_msg_spongos);
+        let (message, spongos) = preparsed.unwrap(branch_announcement).await?;
+
+        let previous_topic = message.payload().content().previous_topic();
+        // Store spongos
+        self.state.spongos_store.insert(address.relative(), spongos);
+        // Store updated cursor for publisher on previous branch
+        self.state.cursor_store.insert_cursor(previous_topic, publisher, cursor);
```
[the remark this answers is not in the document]
DyrellC: It does, and I like that, as is we're pasically just putting that previous topic in twice. I'll make an adjustment

streams/src/api/user.rs
DyrellC: Makes sense to do it now yes.

streams/src/api/user.rs
```diff
@@ -197,21 +200,15 @@ impl<T> User<T> {
...
     /// Sets the latest message link for a specified branch. If the branch does not exist, it is
     /// created
     fn set_latest_link(&mut self, topic: &Topic, latest_link: MsgId) -> Option<InnerCursorStore> {
         self.state.cursor_store.set_latest_link(topic, latest_link)
     }
 
-    fn get_anchor(&self, topic: &Topic) -> Option<MsgId> {
-        self.state.cursor_store.get_anchor(topic)
-    }
+    // fn get_anchor(&self, topic: &Topic) -> Option<MsgId> {
```
kwek20, line 209: remove this :)

streams/src/api/user.rs
```diff
@@ -560,28 +590,59 @@ where
...
+        ensure!(
+            self.transport.recv_message(stream_address).await.is_err(),
+            anyhow!("stream with address '{}' already exists", stream_address)
+        );
+        let send_response = self.transport.send_message(stream_address, transport_msg).await?;
+
+        // If a message has been sent successfully, insert the base branch into store
+        self.state.cursor_store.new_branch(topic.clone());
+        // Commit message to stores
+        self.state
```
kwek20, line 624: 624-630 happen in here, `handle_announcement` and in `handle_branch_announcement`. Maybe we should make a method?

streams/src/api/user.rs
DyrellC: Yeah there are a couple of commented out functions I need to remove entirely, I have already in some changes I have locally but I may need to make a new PR out of that one and add these other changes as a simple commit for this one

streams/src/api/user.rs
```diff
@@ -358,16 +394,10 @@ impl<T> User<T> {
         let author_identifier = self.state.author_identifier.as_ref().ok_or_else(|| {
```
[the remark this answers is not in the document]
DyrellC: I'd rather leave it, we'd have to add it back once we have your PR in since the channel address will be a squeezed representation of the DID based user ids

streams/src/api/user.rs
DyrellC: Wouldn't hurt

streams/src/api/user.rs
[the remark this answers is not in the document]
DyrellC: Ah no I see what you mean, yes I'll remove that.

streams/src/api/user.rs
DyrellC: After examining it more closely I'm not sure that would work. When handling messages we store the cursor before unwrapping the message, and then store spongos if it was successful, whereas on sending we store the cursor and spongos one after another after the message sends properly.

streams/src/api/user.rs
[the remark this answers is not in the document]
DyrellC: After some work/thought I think a different PR makes more sense.

streams/src/api/user.rs
```diff
@@ -266,21 +248,67 @@ impl<T> User<T> {
...
+
+        let new_topic = message.payload().content().new_topic();
+        // Store spongos
+        self.state.spongos_store.insert(address.relative(), spongos);
+        // Insert new branch into store
+        self.state.cursor_store.new_branch(new_topic.clone());
+        // Collect permissions from previous branch and clone them into new branch
+        let prev_permissions = self
+            .cursors()
+            .filter(|cursor| cursor.0 == &prev_topic)
```
arnauorriols, line 302: This iterates the potentially very long set of all cursors in the StreamCan, just to filter those of an specific topic. We could easily have a `(&Topic) -> impl Iterator<Item=(&Identifier, usize)>` method that uses the `HashMap` instead

arnauorriols (approved):
Ok, from my side it's a go! This last comment can be addressed now, or when we implement the advanced message consumption

streams/src/api/user.rs
DyrellC: Alright I added the function, just let me know if there's another change you want done to it, otherwise I'm ready to merge it whenever

kwek20 (approved):
Yes LGTM, lets merge @arnauorriols

PR #263, opened 2022-06-30
implement `uTangle` transport

arnauorriols:
# Description of change

Implement a new `Transport` client for the `Tangle`. The main goal of this `Transport` is to break the hard-dependency on iota.rs, and have an alternative that is lighter, but more importantly, under Streams control. 

# Rationale
iota.rs implements a full-featured client for the Tangle. It is (and remains) the main interface for interacting with the Tangle with Rust. However, it is rather bloated, and it currently has some bugs (for example pow-timeout thread leak) and design decisions (for example dependencies on Bee) that make it less than ideal. Considering the little that Streams actually needs from iota.rs, it is sensible to provide an alternative `Transport` client that is lighter, simpler to maintain, and focused strictly on Streams needs. User can keep opting for the full-featured iota.rs client, but they now have an alternative for which we can better answer for.

## Objectives
- Full implementation control
- Faster & smaller builds
- stabilization of dependency tree 

### Performance
At the time of writting, the new `uTangle` transport has a marginally worse performance than iota.rs transport, for reasons still under investigation (internal implementation is largely verbatim from iota.rs, except for the PoW concurrency handling):
lines

### Bloat
Streams with `uTangle` has a ~1MB smaller code segment, and ~2MB smaller binary. This is including the rayon dependency for the thread-pool. If we wanted to optimize for size, we could replace rayon for the iota.rs concurrency handling for even bigger size reduction.
```
$ cargo bloat -p streams --no-default-features --release --crates --features utangle-client
    Analyzing target/release/libstreams.so

 File  .text     Size Crate
 4.9%  18.1% 916.3KiB std
 4.4%  16.2% 821.7KiB reqwest
 2.5%   9.3% 470.3KiB rustls
 2.3%   8.3% 421.9KiB curve25519_dalek
 1.5%   5.5% 281.0KiB h2
 1.4%   5.1% 261.2KiB tokio
 1.2%   4.3% 220.6KiB ring
 0.9%   3.4% 174.3KiB hyper
 0.9%   3.2% 162.1KiB encoding_rs
 0.5%   1.9%  95.3KiB serde_json
 0.4%   1.6%  83.3KiB url
 0.4%   1.6%  82.1KiB [Unknown]
 0.3%   1.2%  60.7KiB http
 0.3%   1.1%  57.2KiB lets
 0.3%   1.1%  53.9KiB crypto
 0.3%   1.0%  52.8KiB memchr
 0.3%   0.9%  47.3KiB rayon_core
 0.2%   0.7%  36.8KiB ipnet
 0.2%   0.7%  35.8KiB hashbrown
 0.2%   0.7%  33.0KiB idna
 2.8%  10.1% 514.5KiB And 68 more crates. Use -n N to show more.
27.3% 100.0%   5.0MiB .text section size, the file size is 18.2MiB
```
```
cargo bloat -p streams --no-default-features --release --crates --features tangle-client
    Analyzing target/release/libstreams.so

 File  .text      Size Crate
 4.8%  16.8% 1019.0KiB std
 3.9%  13.5%  821.6KiB reqwest
 2.2%   7.7%  469.8KiB rustls
 2.0%   6.9%  421.9KiB curve25519_dalek
 1.3%   4.7%  285.6KiB h2
 1.2%   4.3%  261.5KiB tokio
 1.2%   4.0%  244.7KiB regex_syntax
 1.0%   3.6%  220.6KiB ring
 0.9%   3.1%  186.4KiB iota_client
 0.8%   2.9%  174.3KiB hyper
 0.8%   2.7%  162.1KiB encoding_rs
 0.6%   2.2%  131.0KiB regex
 0.6%   2.0%  121.3KiB time
 0.5%   1.7%  104.6KiB serde_json
 0.4%   1.4%   83.3KiB url
 0.4%   1.3%   82.1KiB [Unknown]
 0.3%   1.2%   71.4KiB bee_message
 0.3%   1.0%   62.8KiB bee_rest_api
 0.3%   1.0%   62.5KiB http
 0.3%   0.9%   53.9KiB crypto
 3.8%  13.3%  808.8KiB And 76 more crates. Use -n N to show more.
28.6% 100.0%    5.9MiB .text section size, the file size is 20.8MiB
```

### Build Time
In a 16-core i7-11800H, Streams with uTangle compiles 4 seconds faster than with Iota.rs. Without rayon the reduction could be significantly bigger. Also keep in mind that in less powerful machines, the difference is bound to be proportionally larger. 
uTangle:
image
Iota.rs
image

The example-tests now include the option to be run with utangle transport (`TRANSPORT=utangle cargo run --example full-example`)

## Change checklist


lets/src/transport/tangle.rs
```diff
@@ -106,36 +105,37 @@ impl TryFrom<IotaMessage> for TransportMessage {
...
-    /// Hash the content of the [`Address`] using `Blake2b256`
-    pub fn to_blake2b(self) -> [u8; 32] {
-        let hasher = Blake2b256::new();
-        hasher.chain(self.base()).chain(self.relative()).finalize().into()
-    }
-
-    /// An `Address` is used as index of the message over the Tangle. For that,
-    /// its content is hashed using [`Address::to_blake2b()`].
-    ///
-    /// ```
```
kwek20, line 119: We should copy over these  comments as well

lets/src/transport/utangle.rs
```diff
@@ -0,0 +1,278 @@
...
+            .send()
+            .await?
+            .json()
+            .await?;
+        Ok(vec![msg.data.try_into()?])
+    }
+}
+
+fn nonce(data: &[u8], target_score: f64) -> Result<u64> {
+    let target_zeros = (((data.len() + NONCE_SIZE) as f64 * target_score).ln() / LN_3).ceil() as usize;
```
kwek20, line 174: we would want to have this as a trait instead, so people can use their own nonce provider, or use the external providers

lets/src/address.rs
```diff
@@ -86,6 +87,16 @@ impl Address {
     pub fn base(self) -> AppAddr {
         self.appaddr
     }
+
+    /// Hash the content of the [`Address`] using `Blake2b256`
+    pub fn to_blake2b(self) -> [u8; 32] {
```
DyrellC, line 92: I like bringing this behaviour into `Address` instead of in `tangle`, seems more appropriate

lets/src/transport/utangle.rs
```diff
@@ -0,0 +1,278 @@
...
+// Local
+use crate::{address::Address, message::TransportMessage, transport::Transport};
+
+const NONCE_SIZE: usize = core::mem::size_of::<u64>();
+// Precomputed natural logarithm of 3 for performance reasons.
+// See https://oeis.org/A002391.
+const LN_3: f64 = 1.098_612_288_668_109;
+
+#[derive(Debug, Clone)]
+pub struct Client<Message = TransportMessage, SendResponse = Ignored> {
```
DyrellC, line 33: Very lean composition :+1: It's nice to see just the essentials.

lets/src/id/identity.rs
```diff
@@ -1,9 +1,10 @@
...
-use alloc::{boxed::Box, string::ToString};
-use core::{
-    convert::{AsRef, TryFrom},
-    hash::Hash,
-};
+use alloc::boxed::Box;
+#[cfg(feature = "did")]
+use alloc::string::ToString;
+#[cfg(features = "did")]
+use core::convert::AsRef;
```
DyrellC, line 6: I don't believe we need this import anymore do we?

streams/examples/full-example/main.rs
```diff
@@ -54,37 +62,85 @@ async fn main_pure() -> Result<()> {
...
+        &node_url
+    );
+    println!(
+        "#####################################################{}",
+        "#".repeat(node_url.len())
+    );
+    Ok(())
+}
+
+#[cfg(feature = "utangle-client")]
```
DyrellC, line 112: I don't think it's necessary for us to be having 2 separate `main_client` implementations. The only real difference in both functions is the `transport` so it would  shave a good number of line changes off to just have something like 
```
    #[cfg(all(feature = "tangle-client", not(feature = "utangle-client")))]
    let transport: Rc<RefCell<tangle::Client>> =
        Rc::new(RefCell::new(tangle::Client::for_node(&node_url).await.unwrap_or_else(
            |e| panic!("error connecting Tangle client to '{}': {}", node_url, e),
        )));
    #[cfg(all(feature = "utangle-client", not(feature = "tangle-client")))]
    let transport: Rc<RefCell<utangle::Client>> = Rc::new(RefCell::new(utangle::Client::new(&node_url)));
```
at line 89 and the function can run as it did before. The extra `#cfg[feature = "utangle/tangle-client"]` lines down in `main()` could be removed, and the messaging wouldn't need to be changed. Unless the intent was to have that differentiation for the benefit of the user, but I'm not sure it would really makes a difference to their interpretation.

streams/src/api/messages.rs
```diff
@@ -53,8 +53,7 @@ use crate::api::{
 /// # let test_transport = Rc::new(RefCell::new(bucket::Client::new()));
 /// #
 /// let author_seed = "cryptographically-secure-random-author-seed";
-/// let author_transport: tangle::Client =
-///     tangle::Client::for_node("https://chrysalis-nodes.iota.org").await?;
+/// let author_transport: tangle::Client = tangle::Client::for_node("https://chrysalis-nodes.iota.org").await?;
```
DyrellC, line 56: I think it would be appropriate to use `utangle::Client` in examples/docs from now on, as it's a leaner client implementation that we should promote more than iota.rs

DyrellC (commented):
Overall I think the implementation looks very nice. Very lean which is nice to see, and should be relatively easy to modify when necessary. I agree with Brord though that it wouldn't hurt to add a little flexability/user choice for things like nonce generation, and the extra function in the examples doesn't really seem necessary to me.

lets/src/id/identity.rs
kwek20: ive cleaned them in my did pr

lets/Cargo.toml
```diff
@@ -14,11 +14,13 @@ readme = "README.md"
 version = "0.1.2"
 
 [features]
-default = ["tangle-client", "did"]
+default = ["utangle-client"]
```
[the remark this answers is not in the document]
arnauorriols: micro-Tangle 😌

PR #264, opened 2022-07-01
make `UserBuilder` infallible

arnauorriols:
# Description of change

Instead of returning a runtime error when the user does not provide a `Transport` in the `UserBuilder`, this PR proposes to make it a compile error.

Additionally, it proposes to use `uTangle` as the default `Transport` if the feature `utangle-client` is enabled. I'm not convinced about this particular decision though, so I'd like to hear your thoughts on it (keep in mind that we can always remove this default and still require the user to explicitly provide a transport statically). In any case, I do think that the `with_default_transport()` method is a failed design so I propose to remove it regardless. 

This PR is temporarily pointing at the `uTangle` branch. Will change it to `v2.0-dev` once either the `uTangle` branch is merged to `v2.0-dev`, or we ditch the idea of defaulting to `uTangle` transport.

### DIFF
--- a/streams/src/api/user_builder.rs
+++ b/streams/src/api/user_builder.rs
@@ -88,8 +88,6 @@ impl<T> UserBuilder<T> {
     /// let psk1 = Psk::from_seed(b"Psk1");
     /// let psk2 = Psk::from_seed(b"Psk2");
     /// let user = User::builder()
-    ///     .with_default_transport::<utangle::Client>()
-    ///     .await?
     ///     .with_psk(psk1.to_pskid(), psk1)
     ///     .with_psk(psk2.to_pskid(), psk2)
     ///     .build();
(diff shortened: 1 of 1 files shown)


DyrellC (commented):
I can get behind this absolutely, although it'll need to wait on the `utangle` review round to wrap up

streams/src/api/user_builder.rs
```diff
@@ -204,29 +190,37 @@ impl<T> UserBuilder<T> {
...
 where
-    Self: Sized,
+    T: for<'a> Transport<'a>,
 {
-    async fn try_default() -> Result<Self>;
+    fn into(self) -> T;
+}
+
+#[cfg(feature = "utangle-client")]
+impl IntoTransport<utangle::Client> for () {
```
kwek20, line 213: i did not know you could do this :o

Issue #260, opened 2022-06-14
topic in `HDF` should be hashed


arnauorriols:
# Description
`BranchAnnouncement`s securely mask the topic of the new branch being created. However, messages published to the branch include the topic in the `HDF`, which is readable by anybody. 

# Solution
`BranchAnnouncement`s and `Keyload`s (assuming #259) carry over the topic of the branch in a secure manner. This means that any subscriber that is supposed to be aware of a topic to which a message is sent, must have already gotten access to the topic plaintext. If the mapping of the cursors of the branches is done with the hash of the topics instead of their plaintext, the `HDF` only need to include such hash.

DyrellC:
The only problem I see with this is that we store cursors based off of `Topic`, so if we don't have the `Topic` in the HDF, the `insert_cursor()` function will have to occur after unwrapping (which can potentially fail). This would change the logic order for handling each of the messages.

arnauorriols:
> If the mapping of the cursors of the branches is done with the hash of the topics instead of their plaintext, the HDF only need to include such hash.

Doesn't this already accounts for that?

kwek20:
Hmm it would also make it hard to join a stream and iterate all topics  to join the ones you want (like we discussed a few weeks ago)

Issue #256, opened 2022-06-04
Add signature in HDF


arnauorriols:
The sequence of a publisher is incremented just by reading the HDF of the message. Therefore, we need a signature that authenticates the HDF part of the message. 

# Open questions
- Why tagged packets if the HDF is signed anyway?
- Can we use a MAC instead of a ed25519 signature?
- why sign the PCF if the HDF is already signed?

Issue #259, opened 2022-06-14
`Keyload` should carry over the Topic of the branch in ciphertext


arnauorriols:
# Description
`BranchAnnouncement`s _announce_ the topic of the new branch. However, `BranchAnnouncement`s mask the topic in order to not reveal it to subscribers that don't have access to the branch from which the branch is being forked. `Keyload`s, however, might include subscribers into the branch down the line, which might not have access to the `BranchAnnouncement` that created the branch. 

# Solution
`Keyload` messages must _mask_ the topic of the branch right after absorbing the encryption key. This way, any new subscriber included into the branch will be able to unmask the topic of the branch.

PR #273, opened 2022-08-05
[V2.0] HDF Update and TopicHash Introduction

DyrellC:
# Description of change
Addresses #260  and #256. Adds a MAC to the HDF for additional verification as well as adding the new `TopicHash` struct. `CursorStore` now maps branches by `TopicHash` instead of `Topic`, and HDF's now contain the `Topic` as `TopicHash` format instead of plaintext. 

## Type of change

- Previous examples continue to work 
- Tests all continue to pass
- fmt/clippy


DyrellC:
Originally the plan was to include #259  in this PR as well, but what I have found is that there is no reasonable automated scenario that a user that would need to read the `Keyload` would be able to sequence to and see that `Keyload` without having known the `Topic` of the branch first, thus defeating the purpose of carrying the `Topic` through the `Keyload` @arnauorriols

DyrellC:
As a consequence of the above stated, it brings into question the need to store cursors by `TopicHash` at all, as the reason for that was to accomodate behavior brought forward by #259.

streams/src/api/user.rs
```diff
@@ -163,7 +166,11 @@ impl<T> User<T> {
     }
 
     pub fn topics(&self) -> impl Iterator<Item = &Topic> + ExactSizeIterator {
-        self.state.cursor_store.topics()
+        self.state.topics.iter()
+    }
+
+    pub(crate) fn topic_by_hash(&self, hash: &TopicHash) -> Option<Topic> {
```
kwek20, line 172: Looking at the code, this didnt need to be cloned really. a reference is enough

kwek20 (commented):
Small change and then good to go!

streams/src/api/user.rs
DyrellC: If we use a reference we need to clone each time anyways to set the latest link for the branch each time a message is handled

PR #277, opened 2022-08-29
[V2.0] Change Identity "to_identifier"

### LINKED ISSUE #265: Convert `Identity::to_identifier(&self) -> Identifier` to `Identity::identifier(&self) -> &Identifier`
Since `Identifier` now contains `String` and no longer implements `Copy`, it should be considered expensive to create. Therefore, instead of creating it every time the user wants the `Identifier` of an `Identity`, the latter should create its `Identifier` at construction and offer it with the getter `Identity::identifier()`

kwek20:
# Description of change

to_identifier is an expensive call now. This change makes Identifier pre-calculated and return a reference instead

Fixes #265

## Type of change

ran the tests and examples

## Change checklist


arnauorriols (approved):
Looks good!

PR #281, opened 2022-09-21
[V2.0] Message Builder

DyrellC:
# Description of change
Adds a `MessageBuilder` tool to `User`, allowing a user to build their message in builder pattern, sending the appropriate type of message dependent on the builder values. The default when instantiating the builder with `user.message()` will be directed at the base branch of the channel as a masked payload tagged packet.

#### Basic Single Branch Style Message (Tagged Packet)
```
user.message()
    .with_payload(raw_data)
    .send()
    .await?;
```

#### Signed Private Message (Signed Packet)
```
user.message()
    .with_topic(branch_topic)
    .with_payload(raw_data)
    .signed()
    .send()
    .await?;
```
#### Unsigned Public Message (Tagged Packet)
```
user.message()
    .with_topic(branch_topic)
    .with_payload(raw_data)
    .public()
    .send()
    .await?;
```

## Type of change

- Unit tests added
- Previous tests continue to pass
- Examples augmented to mix usage of granular and builder methods

### DIFF
--- a/lets/src/message/topic.rs
+++ b/lets/src/message/topic.rs
@@ -25,12 +25,6 @@ impl Topic {
     }
 }
 
-impl From<&Topic> for Topic {
-    fn from(t: &Topic) -> Self {
-        t.clone()
-    }
-}
-
 impl From<&str> for Topic {
     fn from(t: &str) -> Self {
         Self(t.to_string())
--- a/lets/src/message/topic.rs
+++ b/lets/src/message/topic.rs
@@ -25,6 +25,12 @@ impl Topic {
     }
 }
 
+impl From<&Topic> for Topic {
+    fn from(t: &Topic) -> Self {
+        t.clone()
+    }
+}
+
 impl From<&str> for Topic {
     fn from(t: &str) -> Self {
         Self(t.to_string())
--- a/streams/examples/full-example/scenarios/lean.rs
+++ b/streams/examples/full-example/scenarios/lean.rs
@@ -47,36 +47,11 @@ pub(crate) async fn example<SR, T: GenericTransport<SR>>(transport: T, author_se
     lean_subscriber.receive_message(announcement.address()).await?;
 
     println!("author sends a few messages to the base branch");
-    let first_message = author
-        .message()
-        .with_payload(PAYLOAD)
-        .signed()
-        .send()
-        .await?;
-    let _second_message = author
-        .message()
-        .with_payload(PAYLOAD)
-        .signed()
-        .send()
-        .await?;
-    let middle_message = author
-        .message()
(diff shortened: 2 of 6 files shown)


lets/src/message/topic.rs
```diff
@@ -25,6 +25,12 @@ impl Topic {
     }
 }
 
+impl From<&Topic> for Topic {
+    fn from(t: &Topic) -> Self {
```
kwek20, line 33: i dont think we should have this, it kind of implies we can make a Topic form a reference, but were actually just cloning! Why use a hidddent .into() instead of a visible .clone()?

streams/examples/full-example/scenarios/lean.rs
```diff
@@ -48,21 +47,11 @@ pub(crate) async fn example<SR, T: GenericTransport<SR>>(transport: T, author_se
     lean_subscriber.receive_message(announcement.address()).await?;
 
     println!("author sends a few messages to the base branch");
-    let first_message = author
```
kwek20, line 51: nice cleanup!

streams/src/api/message_builder.rs
```diff
@@ -0,0 +1,292 @@
...
+        } else {
+            public = self.payload.as_ref()
+        }
+
+        let user: &mut User<Trans> = self.user;
+        let topic = user.base_branch().clone();
+        if self.signed {
+            user.send_signed_packet(&topic, public, private).await
+        } else {
+            user.send_tagged_packet(self.topic, public, private).await
```
kwek20, line 133: hmm so tagged packets can only be send in the base branch? Why is that? self.topic is already base_branch without setting. no need to restrict it right?

streams/src/api/message_builder.rs
```diff
@@ -0,0 +1,292 @@
...
+        user.create_stream(BASE_BRANCH).await.unwrap();
+
+        user
+    }
+
+    #[tokio::test]
+    async fn make_message() {
+        let topic = "A message topic";
+        let payload = "A payload";
+        let raw_payload = [80_u8, 97, 121, 108, 111, 97, 100];
```
kwek20, line 160: damn i hoped on a sneaky message here!

streams/src/api/message_builder.rs
DyrellC: No you're correct, that wasn't meant to be restrictive, this is a typo/bug, pushing a fix

streams/src/api/message_builder.rs
DyrellC: haha I considered it for sure

kwek20 (approved):
LGTM!

PR #289, opened 2022-10-31
public function

fiNYvo:
to_hex_string() from private to public

# Description of change

change private function to public

## Type of change

Choose a type of change, and delete any options that are not relevant.

## Change checklist


kwek20:
We implement [LowerHex](https://doc.rust-lang.org/std/fmt/trait.LowerHex.html) for this, which gives better control of the output.
Alternatively, you can call hex::encode(appaddr.as_bytes()).

Although, i dont see a reason why not to expose it as well.

